PR #1, opened 2017-01-07
Fix prefix detection when tarball has ./ as first entry

f0:
whan a tarball's first entry is ./ e.g:

tar -tf open-vm-tools-10.1.0-4449150.tar.gz
	./
	./open-vm-tools-10.1.0-4449150/
	./open-vm-tools-10.1.0-4449150/README.md
	./open-vm-tools-10.1.0-4449150/ReleaseNotes.md
	./open-vm-tools-10.1.0-4449150/open-vm-tools/

autospec  can not detect the tarball prefix and
fails with this error (mock build.log)

	Bad %setup option -n: missing argument

This patch fix this by using the second line
and the second split result

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -79,7 +79,7 @@ def build_untar(tarball_path):
     if tarball_contents:
         tarball_prefix = tarball_contents.splitlines()[0].rsplit("/")[0]
         if tarball_prefix == ".":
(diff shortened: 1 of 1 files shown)


matthewrsj:
You are right, this did need to be fixed, and was fixed internally about 4 days ago, sorry for letting the github repo get behind.

f0:
@matthewrsj  ok .... then this can be closed...

PR #3, opened 2017-02-25
Read & write the options.conf and other options files from the target path

bensallen:
Read and write options.conf and other config files from the target path instead of the current working directory. This allows for a cleaner workflow versus having to change directory into the target and specify target to cwd, eg `autospec.py -t .`. 

Defaults to os.getcwd() + "/" + tarball.name + "/" + options.conf, when target isn't specified.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -158,7 +158,7 @@ def main():
 
     config.setup_patterns()
     config.config_file = args.config
-    config.parse_config_files(build.download_path, args.bump)
+    config.parse_config_files(build.download_path, args.bump, tarball.name, args.target)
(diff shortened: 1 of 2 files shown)


bryteise:
I think this looks good.

@bensallen as an aside we have some internal tooling that makes this interaction easier. I'll poke around to see what can be done to release that tool as it should make things easier for folks working with the clearlinux-pkgs repos.

bensallen:
@bryteise I've started creating some additional tooling here as well: https://github.com/bensallen/autospec-common. Thanks.

bryteise:
Oh awesome, we will probably end up with something pretty similar.

PR #5, opened 2017-09-22
Remove config from lib dependencies

bryteise:
The lib submodule requiring the config submodule was an artifact of
how autospec used to work before the stateless implementation in Clear
Linux. config is now mostly used for systemd configuration and unit
files which do not need to be pulled in for library use cases. So to
avoid having lib pull in systemd units that need the bin submodule,
stop having lib require config.


matthewrsj:
+1, I'd like to get those pkg_integrity tests fixed soon.

PR #6, opened 2017-09-22
Remove asciidoc from failed_commands

bryteise:
Asciidoc being in the configure log should not cause it to be added to
the build requires automatically. This is also intended to enable
asciidoc to be removed from use entirely.


matthewrsj:
+1

PR #11, opened 2017-09-26
Update unittests to run all from one command

matthewrsj:
This allows python3 unittest to properly generate the tests when running
all the tests in the tests directory.

Also adds a fix to test_test.py to reset buildpattern.default_pattern
every test. This was a bug introduced by running all tests at once -
this was not being cleared correctly.

Modifies .travis.yml to call make unittests instead of each test target individually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,4 @@ python:
 # command to run tests
 script:
   - make check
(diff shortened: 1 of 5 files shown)


bryteise:
+1

PR #12, opened 2017-09-26
Skip problem tests on Travis CI only

matthewrsj:
These tests regularly fail in Travis for unknown reasons. Skip them
until the tests can be rewritten.


bryteise:
+1 and yaay!

PR #15, opened 2017-09-28
Only add archive basename to upstream

### LINKED ISSUE #14: archives added to upstream file with full URL
Only the basename should be added after the sha1.

matthewrsj:
Fixes #14

Instead of adding the entire URL to the upstream file, only add the
archive basename after the sha1.


matthewrsj:
@ahkok @bryteise

bryteise:
+1 makes sense to me

PR #20, opened 2017-10-03
Add slang config test

bryteise:
Add slang-dev when packages are probing for slang.h.


matthewrsj:
+1 it works.

I think we should try to at least add to the commit message or PR message in the future which package this fixes to make it easier to test. Maybe we need to write a CONTRIBUTING file? We could also add a pull-request template, but that might annoy developers a bit.

PR #21, opened 2017-10-05
Updates to allow autospec to be installable

### LINKED ISSUE #8: Make autospec installable
Autospec should be installable using setuptools instead of existing as a glorified python script at some path on the user's system.

bensallen:
- Rename autospec package to libautospec to avoid name collisions between autospec.py itself and the package.
- Update all import modules in libautospec to use `from libautospec import ...` to provide appropriate name spacing of modules, without having to change existing calls to functions in said modules.
- Update Makefile to set PYTHONPATH to ${CURDIR}/, so new use `from libautospec import ...` is tested and correctly pathed.
- Update all tests for new package libautospec naming.
- Move autospec.py to scripts directory so it is not included in libautospec package.
- Add autospec.py as a script so setuptools will install it to /usr/bin or similar.

Suggest reviewing 81043e2, which was committed before renaming the autospec directory to libautospec. This commit provides actual diffs.

Should close #8.


matthewrsj:
This is a good change, but I need to do some review of some of the other tools that interface with autospec. I'm fairly certain those tools expect to see `autospec/autospec.py` instead of `libautospec/autospec.py`.

That could be updated, but I can't merge a project structure change until it is done.

bensallen:
@matthewrsj Note, this PR moves autospec/autospec.py to scripts/autospec.py.

Do you have Python tooling that imports autospec/autospec.py directly today or are you saying you have hardcoded the path to autospec/autospec.py? For the former, we could split all but main out of scripts/autospec.py and move the functions to libautospec/autospec.py if needed.

matthewrsj:
unfortunately, we have scripts that explicitly run autospec/autospec.py. This is of course bad practice and needs to be changed. I'll have to do that work before I can merge this PR.

bensallen:
Understood. I'll check up on this PR once and while and rebase as needed.

matthewrsj:
@bensallen could you rebase this? I think I've tracked down everything that needs to change.

matthewrsj:
Hmm, one more issue.

As I mentioned we need to be able to run this via a relative path, i.e. `../../autospec/scripts/autospec.py`. While this PR works if the path is correct, it only works if autospec is installed on the system (libautospec exists in the path).

This is an important use case because we do a lot of modification/testing of autospec internally. From a development standpoint it would be much easier to just modify one of those files and have autospec.py pick up that change automatically without having to reinstall on the system.

Is there a way to enable both use cases?

bensallen:
@matthewrsj Rebased, only conflict was in the Makefile flake8 line.

For development one can run `pip install --user --editable .` from the root of a checkout, which will install a wrapper for autospec.py and libautospec into ~/.local/{bin,lib} from your working source directory. This is nicer then adding sys.path code into autospec.py in my opinion. Another alternative is to run `PYTHONPATH=$PWD python3 ./scripts/autospec.py`, similar to what the Makefile is doing.

matthewrsj:
@phmccarty can I get your input on this (since you are working on the tooling)

matthewrsj:
@phmccarty ping. Need your input on how this could integrate with the tooling you are working on.

matthewrsj:
Closing as this is quite out of date now, #8 is already tracking this issue. Need to collaborate with @phmccarty for an implementation that wouldn't break https://github.com/clearlinux/common tooling.

PR #23, opened 2017-10-06
Deprecate use of Python 3.6 style format strings

### LINKED ISSUE #22: Remove Python3.6-style format strings
A lot of environments using autospec are still on Python 3.5 and older, the 3.6-style format strings should be reverted (`f"..."`)

matthewrsj:
Several platforms are using Python 3.5 or lower. Replace the new f"..."
style strings with the traditional "...".format() style strings.

Fixes #22


mrkz:
lgtm

PR #24, opened 2017-10-10
Allow manual override of requires

bryteise:
Allow developers to specify Requres that will not be dropped during
existence detection normally done for auto-detecting runtime
requirements.


matthewrsj:
I'd rather the os-packages list was more comprehensive and included sub-packages, but that's a lot of work when the packages list is published with just the main packages.

+1

PR #25, opened 2017-10-11
Create mock chroot with a unique uniqueext

matthewrsj:
This prevents collisions when two instances are trying to build the same
package in parallel. Unit tests added as well.


autospec/build.py
```diff
@@ -164,6 +165,26 @@ def parse_build_results(filename, returncode, filemanager):
             success = 1
 
 
+def get_uniqueext(dirn, dist, name):
+    """
+    Find a unique name to create mock chroot without reusing an old one
+    """
+    # Default to tarball name
+    resultsdir = os.path.join(dirn, "{}-{}".format(dist, name))
```
bryteise: Thoughts on wrapping os.mkdir in a try and checking for FileExistsError instead? Prevents a race condition I think when it is run at the same time.

autospec/build.py
```diff
@@ -164,6 +165,26 @@ def parse_build_results(filename, returncode, filemanager):
...
+    if not os.path.exists(resultsdir):
+        return name
+
+    # Find a unique extension by checking if it exists in /var/lib/mock
+    # Increment the pathname until an unused path is found
+    resultsdir += "-1"
+    seq = 1
+    while os.path.exists(resultsdir):
```
bryteise: Again I'd attempt os.mkdir here for testing.

autospec/build.py
matthewrsj: Ah, true. Ok.

bryteise: Hrm well the usual operation is leaving the directory behind... Could make the test a little function that would switch between exists and mkdir depending on if --no-cleanup-after is being used? Starting to feel more okay with the race =P.

matthewrsj:
Latest commit adds the `-C/--cleanup` argument to tell mock to clean up the chroots when it finishes. Only in this case is the uniqueext actually guaranteed to be unique.

autospec/build.py
matthewrsj: the problem with the os.mkdir approach is that you need to be root to touch /usr/bin/mock/

matthewrsj:
ok, `mkdir` is used now.

matthewrsj:
*sigh* those failures aren't happening locally. hold on.

matthewrsj:
Had to add `sudo: enabled` to travis.

bryteise:
+1

PR #27, opened 2017-10-13
Only build travis on master branch

matthewrsj:
Let's not build travis every time someone pushes to their personal branch. This PR is kind of a test to make sure Travis will still run for PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,7 @@ python:
 script:
   - make check
   - make unittests
+
+branches:
+  only:
+    - master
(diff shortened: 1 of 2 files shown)


matthewrsj:
Also skip more pkg_integrity tests that rely on external network (can't reach pypi.debian.net right now).

.travis.yml
```diff
@@ -6,3 +6,7 @@ python:
 script:
   - make check
   - make unittests
+
+branches:
+  only:
+    - master
```
phmccarty: This change will affect autospec forks, right? I generally like to see Travis results when I'm working on a topic branch in a fork...

matthewrsj: ah, that's a good point. This was just so I didn't get emails for everyone's travis builds in this repo. I'll have to rethink this :(

matthewrsj:
Ok, I'll deal with it or come up with another solution.

PR #30, opened 2017-10-16
font-adobe name and version detection

### LINKED ISSUE #28: autospec mis-detects version numbers in font packages (75dpi, 100dpi)
Autospec parsed additional numbers from the package names into the version number.

Source URLs:
* https://www.x.org/releases/individual/font/font-adobe-75dpi-1.0.3.tar.bz2
* https://www.x.org/releases/individual/font/font-adobe-100dpi-1.0.3.tar.bz2

autoconfig detected versions as 75.1.0.3 and 100.1.0.3, respectively.

matthewrsj:
Adobe font packages have -nndpi as part of the name. Add tests and a
regular expression to capture these.

Skip some debian pkg_integrity tests since debian is down and we shouldn't be relying on network for unit tests anyways.

Fixes #28

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -232,6 +232,8 @@ def name_and_version(name_arg, version_arg, filemanager):
     pattern_options = [
         r"(.*?)[\-_](v*[0-9]+[a-zalpha\+_spbfourcesigedsvstableP0-9\.\-\~]*)\.orig\.tar",
         r"(.*?)[\-_](v*[0-9]+[alpha\+_spbfourcesigedsvstableP0-9\.\-\~]*)\.src\.(tgz|tar|zip)",
+        # handle font packages with names ending in -nnndpi
+        r"(.*-[0-9]+dpi)[\-_](v*[0-9]+[alpha\+_sbpfourcesigedsvstableP0-9\.\-\~]*)\.(tgz|tar|zip)",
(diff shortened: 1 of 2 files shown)


matthewrsj:
I'm going to remove the test skips and put them in another PR.

autospec/tarball.py
```diff
@@ -232,6 +232,8 @@ def name_and_version(name_arg, version_arg, filemanager):
     pattern_options = [
         r"(.*?)[\-_](v*[0-9]+[a-zalpha\+_spbfourcesigedsvstableP0-9\.\-\~]*)\.orig\.tar",
         r"(.*?)[\-_](v*[0-9]+[alpha\+_spbfourcesigedsvstableP0-9\.\-\~]*)\.src\.(tgz|tar|zip)",
+        # handle font packages with names ending in -nnndpi
+        r"(.*-[0-9]+dpi?)[\-_](v*[0-9]+[alpha\+_sbpfourcesigedsvstableP0-9\.\-\~]*)\.(tgz|tar|zip)",
```
phmccarty: The `dpi?` bit means to match `d`, `p`, and an optional `i` character. That's probably not the intention...

matthewrsj: correct...

matthewrsj: I removed the '?'

phmccarty:
+1

I think those pattern regexps are in need of refinement, but because the new one is in a similar vein to the current ones, I'm okay with this change.

matthewrsj:
https://github.com/clearlinux/autospec/issues/36 opened for regex issues.

PR #33, opened 2017-10-16
Metadata

matthewrsj:
Deprecate writing to a Makefile in the package directory and write to
the options.conf file instead. Any other infrastructure that needs a
Makefile should write it itself.

Instead of requiring the url argument and guessing where to put the
results from an autospec run, read metadata from options.conf. The
'package' section in options.conf defines the package name, url, and
archives when available.

If options.conf is present and there is at least one match between
detected metadata from the source tarball and the metadata in the
configuration, assume that the target directory is the current
directory. This is overridable with the -t/--target option. If nothing
matches the target directory is <current_dir>/<package_name>.

This functionality allows user to run autospec within the target
directory without passing any arguments, including the previously
required "url" argument if the metadata is present.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -24,6 +24,7 @@ import re
 import shutil
 import subprocess
 import pycurl
+import configparser
 
 import build
 import buildpattern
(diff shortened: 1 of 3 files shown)


matthewrsj:
Tests will fail until #30 is merged with the test skips.

autospec/tarball.py
```diff
@@ -451,10 +471,10 @@ def process(url_arg, name_arg, ver_arg, target, archives_arg, filemanager):
     version = ver_arg
     archives = archives_arg
     tarfile = os.path.basename(url_arg)
-    # set gcov file information
-    set_gcov()
     # determine name and version of package
     name_and_version(name_arg, ver_arg, filemanager)
+    # set gcov file information, must be done after name is set
```
bryteise: This isn't immediately obvious to me for why, could be added in the comment here (or in the commit message but here seems easier).

bryteise:
+1 other than my comment nit which might be due to me not seeing something obvious =P.

autospec/tarball.py
matthewrsj: Comment updated. gcov_file variable is set to "<tarball.name>.gcov" if the gcof file exists.

bryteise: Ah okay, the gcov stuff, thanks!

PR #35, opened 2017-10-16
Skip debian pkg_integrity tests in Travis

matthewrsj:
Debian is down again, and we've seen them go down in the past. We
shouldn't be running these tests in Travis.


phmccarty:
+1

Gabib32:
+1

PR #38, opened 2017-10-17
pkg_integrity: handle domain verification for ftp.gnome.org

### LINKED ISSUE #37: look for verification pattern <pkg>-<version>.sha256sum
This is used by some gnome projects (e.g: http://ftp.gnome.org/pub/GNOME/sources/rest/0.8/), but in this particular case, autospec currently looks for
```
rest-0.8.1.tar.xz.asc
rest-0.8.1.tar.xz.sha256
```
where none of those are available, but `rest-0.8.1.sha256sum` is.

mrkz:
Some gnome integrity verification (sha256sum based) do the following
format: `<project>-<version>.sha256sum`. This verification is handled
correctly by autospec for the download.gnome.org domain, but it misses
the ftp.gnome.org domain, failing the verification when using ftp
subdomain.

This commit adds the handling of the ftp.gnome.org subdomain into the
gnome pattern.

Fixes #37.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -235,7 +235,7 @@ def get_signature_url(package_url):
 
 
 def get_hash_url(package_url):
-    if 'download.gnome.org' in package_url:
+    if 'download.gnome.org' in package_url or 'ftp.gnome.org' in package_url:
         return package_url.replace('.tar.xz', '.sha256sum')
     return None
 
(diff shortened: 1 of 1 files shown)


mrkz:
**-1**

We must use `download.gnome.org` instead of the `ftp.gnome.org` sub-domain, so this change is irrelevant now.

PR #39, opened 2017-10-17
Change how to decide whether use sudo or not with mock

cmarcelo:
Instead of peeking at group membership, check if /usr/bin/mock points to consolehelper.

I've kept the "refactor" change in a separate patch, so it is easy to drop it if not desired.

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -35,7 +35,6 @@ must_restart = 0
 base_path = None
 output_path = None
 download_path = None
-mock_cmd = '/usr/bin/mock'
 uniqueext = ''
 
 
@@ -195,19 +194,19 @@ def get_uniqueext(dirn, dist, name):
     return "{}-{}".format(name, seq)
 
 
-def set_mock():
-    global mock_cmd
+def get_mock_cmd():
     # Some distributions (e.g. Fedora) use consolehelper to run mock,
     # while others (e.g. Clear Linux) expect the user run it via sudo.
-    if not (os.path.basename(os.path.realpath('/usr/bin/mock')) == 'consolehelper'):
-        mock_cmd = 'sudo /usr/bin/mock'
+    if os.path.basename(os.path.realpath('/usr/bin/mock')) == 'consolehelper':
+        return '/usr/bin/mock'
+    return 'sudo /usr/bin/mock'
 
 
 def package(filemanager, cleanup=False):
     global round
     global uniqueext
     round = round + 1
-    set_mock()
+    mock_cmd = get_mock_cmd()
     print("Building package " + tarball.name + " round", round)
 
     # determine uniqueext only once
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -23,7 +23,6 @@ import buildreq
 import re
 import tarball
 import os
-import grp
 import shutil
 import subprocess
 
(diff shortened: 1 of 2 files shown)


autospec/build.py
```diff
@@ -198,11 +198,8 @@ def get_uniqueext(dirn, dist, name):
 
 def set_mock():
     global mock_cmd
-    # get group list of current user
-    user_grps = [grp.getgrgid(g).gr_name for g in os.getgroups()]
-    if os.path.exists('/usr/bin/mock'):
-        if 'mock' not in user_grps:
-            mock_cmd = 'sudo /usr/bin/mock'
+    if os.path.basename(os.path.realpath('/usr/bin/mock')) != "consolehelper":
```
bryteise: I think this is a good idea though I'm inclined to say the test should be if it is mock instead of not consolehelper.

cmarcelo: Changed to check for 'mock', and added a short note about consolehelper.
Fixed the unused import and re-pushed.

bryteise:
+1 from me.

autospec/build.py
```diff
@@ -196,20 +194,19 @@ def get_uniqueext(dirn, dist, name):
...
-def set_mock():
-    global mock_cmd
-    # get group list of current user
-    user_grps = [grp.getgrgid(g).gr_name for g in os.getgroups()]
-    if os.path.exists('/usr/bin/mock'):
-        if 'mock' not in user_grps:
-            mock_cmd = 'sudo /usr/bin/mock'
+def get_mock_cmd():
+    if os.path.basename(os.path.realpath('/usr/bin/mock')) == 'mock':
+        return 'sudo /usr/bin/mock'
```
busykai: On a misconfigured system, `util.call()` may get stuck awaiting input. So maybe `sudo` should be called with `-n` (`--non-interactive`) here to make sure it fails right away. Or is the intent is to provide an prompt in such scenario?

cmarcelo: Thanks for the observation (it will be useful for me in another context). I'm not sure what is the intent of autospec here, I've just kept the original behavior.

cmarcelo:
@miguelinux there are two main ways that mock is used: via consolehelper in which mock is a link to consolehelper, that calls the real mock living elsewhere; and via sudo. The former is used by Fedora, Ubuntu, and the latter is used by Clear Linux. WhenThe intention of this patch is that everything using consolehelper, the distros use the "mock" group as a way to identify if the user has permissions.

The patch changes the heuristic of identifying to be whether mock is consolehelper or not, instead of whether the user belongs to the mock group. This should work correct in other distros.

> Could we detect if autospec is running in Clear Linux and then use the sudo helper?

We could, but since the main difference here is "does the distro use consolehelper or not", I'm detecting that instead.

miguelinux:
@cmarcelo What will happend using another distro (not CRL), and the user is not in the mock group?

cmarcelo:
> @cmarcelo What will happend using another distro (not CRL), and the user is not in the mock group?

If the distro uses consolehelper (Fedora, Ubuntu), it will fail as expected.

I'm assuming that in distros with consolehelper, we don't want to be using sudo but follow the instructions (user need to be in mock group to use mock).

miguelinux:
Ok, I see your point.

miguelinux:
+1 from me

autospec/build.py
```diff
@@ -196,20 +194,19 @@ def get_uniqueext(dirn, dist, name):
...
 
 
-def set_mock():
-    global mock_cmd
-    # get group list of current user
-    user_grps = [grp.getgrgid(g).gr_name for g in os.getgroups()]
-    if os.path.exists('/usr/bin/mock'):
-        if 'mock' not in user_grps:
-            mock_cmd = 'sudo /usr/bin/mock'
+def get_mock_cmd():
```
mrkz: Could you please document briefly (e.g: why we do) the difference between just calling `/usr/bin/mock` vs `sudo /usr/bin/mock`.

something like
```
# if using consolehelper (fedora, and other distros), user must be in the mock
# group to gain privileges. On Clearlinux we use 'sudo' to avoid the use of
# consolehelper, so user in mock group is unnecessary
```
would do fine, so for future reference we understand this better just by reading the comment at the top of the function.

matthewrsj:
+1, but I wouldn't mind the comment that @mrkz suggested.

cmarcelo:
Pushed new version with a comment and back to explicitly check for consolehelper, since I think makes clearer to understand our intention.

PR #40, opened 2017-10-17
specfiles: Quote filenames

### LINKED ISSUE #32: Filenames with spaces must be quoted in %files
CHIRP (http://trac.chirp.danplanet.com/chirp_daily/LATEST/chirp-daily-20171013.tar.gz) creates data files with spaces in their names (e.g. `/usr/share/chirp/stock_configs/US Calling Frequencies.csv`). The generated spec file currently does not quote the filename in `%files data` as required by rpmbuild.

bwarden:
In `%files` sections, double-quote all filenames. Use assumption that all
filenames have a leading slash to detect and separate any macro
prefixes.

This should permit filenames with spaces or other special characters
that otherwise trip up rpmbuild

Fixes #32.

Signed-off-by: Brett T. Warden <brett.t.warden@intel.com>

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -315,7 +315,7 @@ class Specfile(object):
         self._write("%defattr(-,root,root,-)\n")
         if "main" in self.packages:
             for filename in sorted(self.packages["main"]):
-                self._write("{}\n".format(filename))
+                self._write("{}\n".format(self.quote_filename(filename)))
 
         for pkg in sorted(self.packages):
             if pkg in ["ignore", "main", "locales"]:
@@ -324,7 +324,7 @@ class Specfile(object):
             self._write("\n%files {}\n".format(pkg))
             self._write("%defattr(-,root,root,-)\n")
             for filename in sorted(self.packages[pkg]):
-                self._write("{}\n".format(filename))
+                self._write("{}\n".format(self.quote_filename(filename)))
 
     def write_lang_files(self):
         """
@@ -1162,3 +1162,31 @@ class Specfile(object):
 
     def _write_strip(self, string):
         self.specfile.write_strip(string)
+
+    def quote_filename(self, filename):
+        """
+        Quotes the filename, if necessary. Identifies and skips any RPM directive prefix.
+        """
+        # Characters that require quoting -- only those with special
+        # meaning in specfiles
+        special_chars = set(" \t")
+        # Build up the output as a string
+        quoted = ''
+        # Capture any directive prefix separately from actual filename
+        #                          (1                   )(3 )
+        directive_re = re.compile("(%\w+(\([^\)]*\))?\s+)(.*)")
+        parts = directive_re.match(filename)
+        if parts:
+            # Add prefix to the output
+            quoted += parts.group(1)
+            # Set the filename to the remaining portion
+            filename = parts.group(3)
+
+        # Now check for special characters
+        if any(c in filename for c in special_chars):
+            # Quote the filename
+            quoted += '"{}"'.format(filename)
+        else:
+            # Add the filename as-is
+            quoted += filename
+        return quoted
--- a/tests/test_specfile.py
+++ b/tests/test_specfile.py
@@ -343,15 +343,21 @@ class TestSpecfileWrite(unittest.TestCase):
         """
         test write_files base test.
         """
-        self.specfile.packages["main"] = ["mainfile1", "mainfile2", "mainfile3"]
+        self.specfile.packages["main"] = ["mainfile1", "/mainfile2", "/mainfile3",
+                "/mainfile 4", "mainfile\t5", "%foo /mainfile6", "%bar /mainfile 7"]
         self.specfile.packages["ignore"] = ["ignorepkg"]
         self.specfile.packages["other"] = ["other2", "other1"]
         self.specfile.write_files()
+        # Note the special sorting
         expect = ["\n%files\n",
                   "%defattr(-,root,root,-)\n",
+                  '%bar "/mainfile 7"\n',
+                  "%foo /mainfile6\n",
+                  '"/mainfile 4"\n',
+                  "/mainfile2\n",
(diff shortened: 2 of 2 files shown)


bwarden:
Hmmm... maybe `str.split()` is sufficient.

bwarden:
split("/", 1) behaves differently between Python versions when the left side is empty. I'll rework this tomorrow.

phmccarty:
Should autospec be quoting *all* filenames? My preference would be to only quote filenames containing special characters that require quoting.

bwarden:
The problem is, how do you determine which ones need to be quoted? I _could_ just quote any filenames with spaces in them.

matthewrsj:
@phmccarty why not quote all?

phmccarty:
@matthewrsj Personal preference... I think quoting all filenames reduces readability of spec files. I've tried this in the past...

matthewrsj:
Thinking about the diffs of the spec files now... I agree with @phmccarty

bwarden:
Alright, this version only quotes filenames that contain a space.

autospec/specfiles.py
```diff
@@ -324,7 +342,24 @@ def write_files(self):
...
+                    print_fatal("Filenames must start with a / ({0})".format(filename))
+                    sys.exit(1)
+                if len(line) > 1:
+                    # Write the prefix unmodified
+                    self._write("{}".format(line[0]))
+                    # Filename is remaining part
+                    filename = line[1]
+                else:
+                    filename = line[0]
+                if ' ' in filename:
```
matthewrsj: Need to handle any whitespace, including tabs

bwarden: Will fix

bwarden:
Updated tests not only to form correct specfiles, but added patterns for filenames with spaces/tabs, and with rpm directives. Pulled quoting into a separate function that's called from the two places that emit %files.

autospec/specfiles.py
```diff
@@ -1162,3 +1163,35 @@ def _write(self, string):
...
+            # Keep the prefix
+            quoted += line[0]
+            # Filename is remaining part
+            filename = line[1]
+        else:
+            filename = line[0]
+        # Now check for special characters
+        if any(c in filename for c in special_chars):
+            # Quote the remaining filename, with leading slash
+            quoted += "\"/{}\"".format(filename)
```
matthewrsj: Instead of escaping the quotes with `"\"...\""` I would prefer to use single quotes, like this:
```
quoted += '"/{}"'.format(filename)
```
since I think it is easier to read. Same goes for the tests below.

bwarden: Agreed; fixing.

tests/test_specfile.py
```diff
@@ -371,11 +377,11 @@ def test_write_lang_files_base(self):
...
         write_lang_files base test
         """
-        self.specfile.locales = ["pkg1", "pkg2"]
+        self.specfile.locales = ["/pkg1", "/pkg2"]
         self.specfile.write_lang_files()
         expect = ["\n%files locales",
-                  " -f pkg1.lang",
-                  " -f pkg2.lang",
+                  " -f /pkg1.lang",
+                  " -f /pkg2.lang",
```
phmccarty: So, the -f option is special, since it directs rpm to read contents from a file, and the file usually lives in the build directory. Leaving off the leading slash is correct here.

bwarden: Fixing the test.

phmccarty: Great, that looks better, thanks.

autospec/specfiles.py
```diff
@@ -1162,3 +1163,35 @@ def _write(self, string):
...
+        """
+        # Characters that require quoting -- only those with special
+        # meaning in specfiles
+        special_chars = set(" \t")
+        # Build up the output as a string
+        quoted = ''
+        # Separate any directive prefix
+        # This will break if your directive contains a slash!
+        line = filename.split("/", 1)
+        if len(line) < 1:
```
phmccarty: Can `len(line)` ever return 0 in this case? It just tried a quick test in the REPL, and when the filename contains no slashes, the array length is 1.

autospec/specfiles.py
```diff
@@ -1162,3 +1163,35 @@ def _write(self, string):
...
+            print_fatal("Filenames must start with a / ({0})".format(filename))
+            sys.exit(1)
+        elif len(line) > 1:
+            # Keep the prefix
+            quoted += line[0]
+            # Filename is remaining part
+            filename = line[1]
+        else:
```
matthewrsj: I'm concerned about this logic.
```
len("/some/filename".split("/", 1))  # -> 2, hits the elif and quoted contains an empty string
len("%doc /some/filename".split("/", 1))  # -> 2, hits the elif and quoted contains %doc
len("some/filename".split("/", 1))  # -> 2, hits the elif, quoted contains empty string **but** there is no leading /
len("filename".split("/", 1))  # -> 1, hits the else, filname gets the correct value.
len("".split("/", 1))  # -> 1, hits the else, filename is empty, but looks like filename turns into "/" below...
```
When would `len(line)` ever be < 1? It doesn't look like this is actually testing that the file starts with a "/"

autospec/specfiles.py
```diff
@@ -1162,3 +1162,31 @@ def _write(self, string):
...
+        """
+        Quotes the filename, if necessary. Identifies and skips any RPM directive prefix.
+        """
+        # Characters that require quoting -- only those with special
+        # meaning in specfiles
+        special_chars = set(" \t")
+        # Build up the output as a string
+        quoted = ''
+        # Capture any directive prefix separately from actual filename
+        #                          (1                   )(3 )
```
matthewrsj: this was actually a very useful comment.

matthewrsj:
+1!

matthewrsj:
@phmccarty any comments?

phmccarty:
Looks good to me. +1

PR #41, opened 2017-10-18
keyring: add TingPing key for hexchat package sig.

mrkz:


matthewrsj:
+1

PR #42, opened 2017-10-18
Version and Name Detection Enhancements

matthewrsj:
Add the world to the packageurls tests.

Fix up several regular expressions and remove unused ones. Move as much
string replacing to string.replace() functions instead of slow regular
expressions.

Remove the pypi.python.org -> pypi.debian.net url conversion
(pypi.debian.net keeps going down).


bryteise:
+1

autospec/tarball.py
```diff
@@ -249,113 +295,94 @@ def name_and_version(name_arg, version_arg, filemanager):
...
                 version = convert_version(m.group(2))
                 break
 
-    if url.find("bitbucket.org") > 0:
-        p = re.compile(r"https?://bitbucket.org/.*/(.*?)/.*/([.0-9a-zA-Z_-]*?).tar")
-        m = p.search(url)
+    if "mirrors.kernel.org" in url:
+        m = re.search(r".*/sourceware/(.*?)/releases/(.*?).tgz", url)
         if m:
+            print('hit pattern kernel')
```
phmccarty: debug print, I assume?

PR #43, opened 2017-10-20
Add --prep-only option to run preparatory work only

matthewrsj:
For non-autospec enabled packages this option can be used to download
the upstream tarball, any specified archives, extract them and put the
archives at their destination, and update the upstream file, but not
actually attempt to build a specfile.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -116,6 +116,28 @@ def read_old_metadata():
             archives)
 
 
+def write_prep(workingdir):
+    """
+    Write metadata to the local workingdir when --prep-only is used
+    """
+    if config.urlban:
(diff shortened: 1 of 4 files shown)


autospec/autospec.py
```diff
@@ -175,6 +178,14 @@ def main():
             "-a/--archives or options.conf['package']['archives'] requires an "
             "even number of arguments"))
 
+    if args.prep_only:
+        package(args, url, name, archives, "./workingdir")
```
phmccarty: I see an empty directory called "output" within workingdir. Is that directory supposed to be empty?

matthewrsj: I've removed the unused output directory

PR #46, opened 2017-10-21
  support git shortlog from upstream git for creating commit messages

fenrus75:


autospec/commitmessage.py
```diff
@@ -168,6 +170,43 @@ def process_NEWS(newsfile):
...
+        return ""
+    if oldversion == newversion:
+        return ""
+
+    run(["git", "-C", "results", "clone", giturl])
+    p = run(["git", "-C", "results/" + tarball.name, "tag"], stdout=PIPE)
+    tags = p.stdout.decode('utf-8').split('\n')
+
+    for t in tags:
+        i = t.find(oldversion)
```
bryteise: This seems like it could introduce a lot of noise for projects that do version as 1, 2, etc for their versions.

fenrus75: is there a "ends-width" thing in python then?

bryteise:
"string-suffix".endswith("suffix") is what you want.

matthewrsj:
+1

PR #47, opened 2017-10-23
Ignore bare excepts with flake8

matthewrsj:
The previous ignore, import not at top of file (E402) was unneeded. Bare
excepts are handy when we want to fail gracefully from any error. This
is a script, not an imported library.


matthewrsj:
travisfail


wat

matthewrsj:
`¯\_(ツ)_/¯`

bryteise:
+1 and err thanks travis.

PR #50, opened 2017-10-24
Match additional github URL patterns

bwarden:
Extend existing github_patterns to match, e.g.:
https://github.com/Linux-Comedi/comedilib/releases/download/r0_11_0/comedilib-0.11.0.tar.gz

Signed-off-by: Brett T. Warden <brett.t.warden@intel.com>

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -340,7 +340,12 @@ def name_and_version(name_arg, version_arg, filemanager):
             m = re.search(pattern, url)
             if m:
                 repo = m.group(2).strip()
-                name = repo
+                if repo not in name:
+                    # Only take the repo name as the package name if it's more descriptive
+                    name = repo
+                elif name != repo:
+                    name = re.sub("release-", '', name)
+                    name = re.sub("\d*$", '', name)
                 rawname = name
                 version = convert_version(m.group(3))
                 giturl = "https://github.com/" + m.group(1).strip() + "/" + repo + ".git"
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -284,6 +284,7 @@ def name_and_version(name_arg, version_arg, filemanager):
     global version
     global url
     global giturl
+    global repo
 
     tarfile = os.path.basename(url)
 
@@ -338,10 +339,11 @@ def name_and_version(name_arg, version_arg, filemanager):
         for pattern in github_patterns:
             m = re.search(pattern, url)
             if m:
-                name = m.group(2).strip()
+                repo = m.group(2).strip()
+                name = repo
                 rawname = name
                 version = convert_version(m.group(3))
-                giturl = "https://github.com/" + m.group(1).strip() + "/" + name + ".git"
+                giturl = "https://github.com/" + m.group(1).strip() + "/" + repo + ".git"
                 break
 
     if "mirrors.kernel.org" in url:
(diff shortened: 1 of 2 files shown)


matthewrsj:
Please add the github url you referenced to `tests/packageurls`

matthewrsj:
^^ In order sorted by expected package name
```
<url>,<expected-name>,<expected-version>
```

bwarden:
Oh, the URL was already in packageurls. I'll revert the second commit.

matthewrsj:
So, this was already correctly detecting the version then?

bwarden:
Version, yes, but not generating giturl.

matthewrsj:
Oh I see, do you mind adding a couple unit tests then to cover the two main types of github urls?

bwarden:
I read the unit test failures backward and though I was improving things... reworking.

bwarden:
Close review appreciated. There's a little bit of hackishness in here to deal with "release-qpdf" and "libnl3".

autospec/tarball.py
```diff
@@ -332,15 +333,22 @@ def name_and_version(name_arg, version_arg, filemanager):
...
-                           r"https?://github.com/(.*)/(.*?)/releases/download/v.*/(.*).tar"]
+                           r"https?://github.com/(.*)/(.*?)/releases/download/.*?/(.*).tar",
+                           r"https?://github.com/(.*)/(.*?)/files/.*?/(.*).tar"]
 
         for pattern in github_patterns:
             m = re.search(pattern, url)
             if m:
-                name = m.group(2).strip()
+                repo = m.group(2).strip()
+                if repo not in name:
```
matthewrsj: How do we know this is more descriptive if it is different?

tests/test_tarball.py
```diff
@@ -27,10 +28,15 @@ def test_packageurl(self, mock_build):
...
         tarball.url = url
         tarball.name_and_version('', '', FileManager())
-        self.assertEqual(tarball.name, name)
-        self.assertEqual(tarball.version, version)
+        self.assertEqual(name, tarball.name)
+        self.assertEqual(version, tarball.version)
+        if re.match("https?://github.com", url) != None:
+            self.assertIsNotNone(
+                    re.match("https://github.com/[^/]+/"+tarball.repo+".git",
+                    tarball.giturl))
```
matthewrsj: this is asserting that `re.match(...)` is not none, and displaying the `tarball.giturl`as the error message. Was this intended? Seems like we should be checking at least if `tarball.giturl` is not None.

autospec/tarball.py
bwarden: It's only an assumption, but it holds for the current `test/packageurls`.

tests/test_tarball.py
bwarden: tarball.giturl is inside re.match(), so it's testing that string against the regex and checking that the match status is not None.
Nonetheless, I'll add self.assertIsNotNone(tarball.giturl) prior to that.

autospec/tarball.py
bwarden: More specifically, it's a replacement for the prior behavior of _always_ picking the repo name over the tarball-derived name.

tests/test_tarball.py
matthewrsj: ah, /me having `(())` dyslexia

matthewrsj:
+1

PR #51, opened 2017-10-26
Remove empty download file on pycurl failure

### LINKED ISSUE #26: Source tarball download failure creates empty file
If autospec tries to download a source tarball, but the download fails, it creates an empty file. For example, I am currently unable to use it for downloading https://pypi.debian.net/awscli/awscli-1.11.168.tar.gz, resulting in a empty file being created named 'awscli-1.11.168.tar.gz'.

matthewrsj:
Fixes #26

Remove empty file created when a download fails. The empty file caused
autospec to fail on consecutive runs due to autospec attempting to reuse
the file when it exists.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -66,6 +66,7 @@ def really_download(upstream_url, destination):
             c.perform()
         except pycurl.error:
             print_fatal("unable to download {}".format(upstream_url))
+            os.remove(destination)
             exit(1)
         finally:
             c.close()


matthewrsj:
@phmccarty I was unable to reproduce any issue with signature files hanging around. If you can find an example of this happening let me know.

phmccarty:
@matthewrsj I cannot find the package where I ran into that issue, but I'll report a new bug if I see it again.

phmccarty:
+1

PR #54, opened 2017-10-27
Ensure R packages contain full debuginfo

phmccarty:
Because the -g option is not specified by default when linking R shared libraries, add it manually to the PKG_LIBS variable in Makevars to inject -g into the linker command line. This ensures that the shared library (or libraries) contain full debuginfo.


phmccarty:
Needs more work

Issue #31, opened 2017-10-16
Remove network dependency from pkg_integrity tests 


matthewrsj:
The pkg_integrity unit tests rely on network connection, making them brittle (especially on Travis). These tests should mock their network connection instead.

matthewrsj:
Fixed by #55

PR #55, opened 2017-11-01
Fix Issue #31: Remove network dependency from pkg_integrity tests

Gabib32:
This pull request uses two mock functions to imitate downloading files, and getting the status of a head request. This removes the network calls and allows tests to be ran in Travis CI.


tests/test_pkg_integrity.py
```diff
@@ -1,50 +1,31 @@
 import os
+import shutil
+import mock
 import unittest
 import tempfile
 import pkg_integrity
 
+BASEDIR = os.getcwd()
+TESTDIR = BASEDIR + '/tests/testfiles/pkg_integrity'
```
bryteise: os.path.join() is your friend (though we don't do that as much as we should in autospec).

bryteise:
Looks good minus my nit. @Gabib32 is there a reason to not have these be all in one commit?

tests/test_pkg_integrity.py
matthewrsj: It's been getting better! Not all the way better, but better.

matthewrsj: It doesn't look like BASEDIR gets used anywhere, is that the case? If so, can you just do 
```TESTDIR = os.path.join(os.getcwd(), 'tests/testfiles/pkg_integrity')```

matthewrsj:
Other than the little thing +1 from me as well. Nice use of mock.patch.

Gabib32:
Made those changes, and learned about squashing!

Issue #52, opened 2017-10-27
directories included in file list


matthewrsj:
Occasionally (and unpredictably and un-reproducibly) we see directories being added to the file list.

phmccarty:
A patch was added to the Clear Linux 'rpm' package recently that was intended to address this issue.

https://github.com/clearlinux-pkgs/rpm/commit/cd19a99148b64e5123207194a2dcf3c1186aa3ed

matthewrsj:
That patch is a guess  to solve a transient problem, I still think autospec could be excluding directories. I'll reply in #58 to your comment about `%dir` files.

matthewrsj:
Fixed by #58

PR #58, opened 2017-11-07
Remove directories from file list

matthewrsj:
Clean directories from package file lists. If directories are
encountered, print a warning, add the directory to the blacklist, and
re-run.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -270,6 +270,15 @@ def package(args, url, name, archives, workingdir):
         filemanager.load_specfile(specfile)
         specfile.write_spec(build.download_path)
         filemanager.newfiles_printed = 0
+        mock_chroot = "/var/lib/mock/clear-{}/root/builddir/build/BUILDROOT/" \
+                      "{}-{}-{}.x86_64".format(build.uniqueext,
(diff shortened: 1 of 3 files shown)


matthewrsj:
Fixes #52

phmccarty:
I'm not sure I like this change. Autospec currently doesn't support installing empty directories (with `%dir /path/to/dir`), but what if it does in the future? For example, in Clear Linux, the filesystem package is not currently autospec'd but mostly it installs empty directories.

matthewrsj:
Like you said, autospec does not currently support installing empty directories.

If it does support it in the future it is pretty simple to skip those files that are prefixed with `"%dir "` when cleaning the directories.

phmccarty:
Okay, then I would recommend adding a large FIXME in the code to explain that the code needs adjustment if autospec starts to support `%dir`.

matthewrsj:
OK, I'll add that.

matthewrsj:
Better yet, I will skip processing of `%dir` files right now anyways, so nothing will need to be changed in the future if that becomes supported.

phmccarty:
Looks okay. +1

Issue #45, opened 2017-10-20
Archive logs per build round


phmccarty:
Instead of overwriting mock's logs for each build round, it would be nice for the logs from each round to be archived for later inspection.

matthewrsj:
Fixed by #59

PR #59, opened 2017-11-07
Save build.log files from failed rounds

matthewrsj:
This is useful for debugging purposes. The file must be saved to the
target directory since the results directory is wiped on each round.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -19,6 +19,7 @@
 import argparse
 import sys
 import os
+import shutil
 import re
 import tempfile
 import configparser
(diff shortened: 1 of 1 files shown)


matthewrsj:
Fixes #45

autospec/autospec.py
```diff
@@ -116,6 +117,17 @@ def read_old_metadata():
             archives)
 
 
+def save_build_log(path, iteration):
+    """
+    Save build log to <path>/results/build.log.<iteration>
+
+    Must be saved outside of the results/ directory since it gets wiped away on
+    each round.
```
phmccarty: It looks like the log is saved to a new file within the results directory, if I'm reading the code correctly? So is this comment out of date?

matthewrsj: Out of date comment.

phmccarty:
+1

autospec/autospec.py
matthewrsj: Resolved. And actually you were mistaken about the code. The comment on line 122 was out of date, not the comment on 124. The `<path>/results/build.log` is copied to `<path>/build.log.round<iteration>`

phmccarty: Ah, right. I think I did misread the code. Thanks.

PR #60, opened 2017-11-07
catkin: introduce catkin pattern recognition

dorileo:
Catkin is based on cmake and is a way to package/build modules as a
cmake components. This patch recognizes the patterns and handle the
dependencies, every catkin package provides a package-config descriptor
so it's pretty safe to rely on that.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -33,6 +33,7 @@ import subprocess
 banned_requires = set()
 buildreqs = set()
 requires = set()
+extra_cmake = set()
 verbose = False
 cargo_bin = False
 banned_buildreqs = set(["llvm-devel",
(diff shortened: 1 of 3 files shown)


matthewrsj:
Can you provide the package URL you were testing with?

dorileo:
@matthewrsj  I've recently tested these (there're more but I think this is good enough for now):

https://github.com/Eurecat/abseil-cpp/archive/0.2.0.tar.gz
https://github.com/ros/console_bridge/archive/0.3.2.tar.gz
https://github.com/ros/gencpp/archive/0.5.5.tar.gz
https://github.com/jsk-ros-pkg/geneus/archive/2.2.6.tar.gz
https://github.com/ros/genlisp/archive/0.4.16.tar.gz
https://github.com/ros/genmsg/archive/0.5.9.tar.gz
https://github.com/RethinkRobotics-opensource/gennodejs/archive/2.0.1.tar.gz
https://github.com/ros/genpy/archive/0.6.7.tar.gz
https://github.com/ros/message_generation/archive/0.4.0.tar.gz
https://github.com/ros/std_msgs/archive/0.5.11.tar.gz

autospec/buildreq.py
```diff
@@ -560,3 +585,4 @@ def load_specfile(specfile):
     specfile.buildreqs = buildreqs
     specfile.requires = requires
     specfile.cargo_bin = cargo_bin
+    specfile.extra_cmake = " ".join(extra_cmake)
```
matthewrsj: It looks like this overwrites the `extra_cmake` args specified via the `cmake_args` configuration file (https://github.com/clearlinux/autospec/blob/master/autospec/config.py#L672).

Why not just append like so:
```
specfile.extra_cmake += " " + " ".join(extra_cmake)
```
so user-set cmake args aren't being overwritten.

dorileo: done

matthewrsj: The change still won't add a new space between the previous `extra_cmake` and the new `extra_cmake`, so potentially you would have `-USER -FLAGS-CATKIN -FLAGS` when you should have a space between `-FLAGS` and `-CATKIN`.

matthewrsj:
My old comment was collapsed because of the force push. Here it is again:

The change still won't add a new space between the previous `extra_cmake` and the new `extra_cmake`, so potentially you would have `-USER -FLAGS-CATKIN -FLAGS` when you should have a space between `-FLAGS` and `-CATKIN`.

dorileo:
> My old comment was collapsed because of the force push. Here it is again:
> The change still won't add a new space between the previous extra_cmake and the new extra_cmake, so potentially you would have -USER -FLAGS-CATKIN -FLAGS when you should have a space between -FLAGS and -CATKIN.

[Done] thanks.

matthewrsj:
+1, thank you.

PR #62, opened 2017-11-07
Respect urlban when writing metadata to options.conf

matthewrsj:
This also requires moving the read_config_opts() call to after
autospec.conf is read so the urlban is populated.


phmccarty:
+1

PR #63, opened 2017-11-07
license: Add tests for complete coverage of license module

matthewrsj:
The additional test uncovered a syntax error in the print_fatal call in
license_from_copying_hash, which was fixed as well. An unnecessary mock
was removed from another test.


bryteise:
+1

PR #64, opened 2017-11-08
Increase count.py test coverage

matthewrsj:
Also remove some never-hit regular expressions from count.py


bryteise:
+1

PR #66, opened 2017-11-09
Add efivar and gnu-efi simple patterns

bryteise:
Enable detecting missing efivar and gnu-efi build requirements (with
meson).


matthewrsj:
+1

Issue #29, opened 2017-10-14
Stop writing metadata to a Makefile


matthewrsj:
This is not needed by autospec. The URL and archives should be part of autospec per-package configurations instead.

matthewrsj:
Fixed by #67

PR #67, opened 2017-11-13
Deprecate Makefile writing

matthewrsj:
The package metadata is now written to options.conf so the Makefile
writing is no longer needed.


bryteise:
+1

phmccarty:
+1

matthewrsj:
Fixes #29

PR #68, opened 2017-11-15
Infile parser to use bitbake files

Gabib32:
The first phase of the infile parser is complete. When the --infile arg is passed with a url, file, or directory of files that is a recipe file it scrapes the information and stores it to update the specfile. If the infile arg is passed a url is not needed, and if one is not present the scraped tarball from the .bb file will be used. Additionally, any licenses taken from the recipe file will be added to the licenses list, thus preventing autospec failures that cannot find a .license file. The summary and build deps are added to the specfile after the initial creation of the specfile by autospec.


bryteise:
I'm pretty sure the last three commits need to have a different entity as the committer. I'd also like to see the fixes squashed into the commits that they would have fallen into normally instead of added on to the series.

matthewrsj:
It looks like there are several "fix" commits sprinkled through there as well. Those should be squashed into the commits they are fixing as well.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+    # remove line continuations
+    value = "".join(value.split("\\"))
+    return value
+
+
+def update_inherit(line, bb_dict):
+    if 'inherits' in bb_dict:
+        bb_dict['inherits'].append(' '.join(line.split(' ', 1)[1:]))
+    else:
+        bb_dict['inherits'] = line.split(' ', 1)[1:]
```
matthewrsj: if you follow my above suggestion this does not have to be updated (since you are expecting it to be a list) but it might be a little less confusing if you just stored it as an initiated list like this:
`bb_dict['inherits'] = [line.split(maxsplit=1)[1]]`

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+    return depth
+
+
+def pattern_match_line(line):
+
+    operators = ["??=", "?=", ":=", "+=",
+                 "=+", ".=", "=.", "="]
+
+    for i, e in enumerate(operators):
+        operators[i] = '\\' + '\\'.join(e)
```
matthewrsj: this logic produces:
`['\\?\\?\\=', '\\?\\=', '\\:\\=', '\\+\\=', '\\=\\+', '\\.\\=', '\\=\\.', '\\=']`
Why the double backslash? Why not just define the list like that in the first place? It makes it easier to understand what is going on.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+                        break
+                    else:
+                        line = next(bb_fp).strip()
+
+                if cmd[1].startswith("# " + cmd_name) and cmd[-1] == '# }':
+                    cmd = cmd[2:-1]
+
+                bb_dict[cmd_name] = cmd
+                continue
+            #
```
matthewrsj: Extra `#`

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+            bb_dict[key] = evaluate_expr(op, bb_dict.get(key), value)
+            if op == "??=":
+                bb_dict[key] = [bb_dict[key], 1]
+        except AttributeError as error:
+            print("Missing operation functionality: {}".format(error))
+
+    return bb_dict
+
+
+def bb_scraper(bb_fp, bb_dict):
```
matthewrsj: It would be much simpler to pass in a list of lines to this function instead of the open file handler. Then instead of calling `next()` in an infinite `while` loop you can iterate on the line list. This would simplify a lot of your code.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+        bb_dict = scrape_version(bb_dict)
+
+    for line in bb_fp:
+        cont = None
+        if line.strip() and not line.strip().startswith('#'):
+            line = line.strip()
+
+            # If line is a command, create raw string of the command
+            # TODO: command could be python code
+            if not cont and line.startswith('do_'):
```
matthewrsj: I don't see where you ever set `cont` to anything but `None`, so that variable is basically a no-op.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+            elif not cont and line[-1] == '\\':
+                lines = []
+                while 1:
+                    lines.append(line)
+                    line = next(bb_fp).strip()
+                    if line[-1] != '\\':
+                        lines.append(line)
+                        break
+
+                line = " ".join(lines)
```
matthewrsj: Above you say `src_uri line continuation should be a list, not a string`, but then you join the list into a string here.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+
+            if line.startswith('inherit'):
+                update_inherit(line, bb_dict)
+                continue
+
+            match = pattern_match_line(line)
+            if match:
+                bb_dict = write_to_dict(bb_dict, match)
+            else:
+                todo.append(line)
```
matthewrsj: Looks like you append to `todo` but never use it, is this just for debugging?

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def parse_ext(path):
+    ext = os.path.splitext(path)[1]
+    if ext not in parsable_filetypes:
+        util.print_warning("Cannot parse infile \"{}\" type. "
+                           "from input: {}".format(ext, path))
+        return
+
+    if ext.startswith('.'):
```
matthewrsj:
```
ext = ext.lstrip('.')
```

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+        ext = ext[1:]
+
+    return ext
+
+
+def check_url_content(url):
+    if "text/html" in requests.head(url).headers['content-type']:
+        if 'git' in url and "/tree/" in url:
+            url = url.replace("/tree/", "/plain/", 1)
```
matthewrsj: is this url scheme (tree -> plain) valid for all git hosting services, or just for yoctoproject? It looks like this doesn't work for github.

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+                             "please use plaintext.")
+            return None
+    return url
+
+
+def sort_files(x):
+    return parsable_filetypes.get(os.path.splitext(x)[1])
+
+
+def parser_type(bb_fp, output_dict, parse_type):
```
matthewrsj: This returns the dict result of the parse, not the type of parser. Consider renaming.

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+    parse_type = parse_ext(indata)
+
+    if output_dict.get('filename'):
+        output_dict['filename'].append(indata)
+    else:
+        output_dict['filename'] = [indata]
+
+    if not os.path.isfile(indata):
+        # check that input is plain or raw text and not html
+        indata = check_url_content(indata)
```
matthewrsj: `indata` might be set to `None` by `check_url_content` if the url is not for a plaintext file. What happens in `urlretrieve` if this happens?

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            try:
+                tmp, _ = urlretrieve(indata, tmpfile.name)
+                with open(tmp, 'r') as bb_fp:
+                    output_dict = parser_type(bb_fp, output_dict, parse_type)
+            except Exception as e:
+                util.print_warning("Error downloading url: {}".format(e))
```
matthewrsj: Ah, this is what happens. Is this designed to continue the autospec run if the `infile` is invalid?

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+      A directory with multiple files or urls
+      A path or filename
+
+    Each file in the directory should scraped to the same dictionary instance.
+    """
+    output_dict = {"NAME": name}
+
+    if os.path.isdir(indata):
+        files = [f for f in os.listdir(indata) if os.path.isfile(os.path.join(indata, f))]
+        for file in sorted(files, key=sort_files):
```
matthewrsj: try not to shadow the `file` keyword.

autospec/infile_handler.py
```diff
@@ -0,0 +1,121 @@
...
+            print(file)
```
matthewrsj: debug print?

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,75 @@
...
+
+def update_summary(bb_dict, specfile):
+    """
+    Updates the default summary to the summary or description scraped from
+    the bitbake file IF the default summary is "No detailed summary available"
+    """
+    k = "default_sum"
+    v = "No detailed summary available"
+
+    if hasattr(specfile, k) and getattr(specfile, k) == v:
```
matthewrsj: You shouldn't need to check for existence of these attrs, we know specfile has them because they are defined to have them.

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,75 @@
...
+        for i in ['SUMMARY', 'DESCRIPTION']:
```
matthewrsj: `i` usually indicates an index, can you name this something more descriptive?

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,75 @@
...
+            if i in bb_dict:
+                setattr(specfile, k, bb_dict.get(i))
```
matthewrsj: why not just `specfile.default_sum = bb_dict.get(i)`? Have to be careful though because `.get()` can return `None` and you should default `default_sum` to `""`

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,75 @@
...
+                break
+
+
+def update_licenses(bb_dict, specfile):
+    """
+    The specfile contains a list of licenses for a package, if the bitbake
+    license is not included in that list, add it.
+    """
+    if hasattr(specfile, "licenses") and "LICENSE" in bb_dict:
```
matthewrsj: Same comment as above about using hasattr when we know it has it.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+    "+=": "{0} {1}",
+    "??=": "{1}",
+    "?=": "{1}"
+}
+
+
+def scrape_version(bb_dict):
+    # get version from latest filename added to dict, if exists
+    try:
+        bb_dict['VERSION'] = bb_dict.get('filename')[-1].split('_', 1,)[1].rsplit('.', 1)[0]
```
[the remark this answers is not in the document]
Gabib32: This is the version of the bitbake file package. It is used to replace the variable ${PV} within the bitbake file. That does include replacement in the URL that is used only if one is not passed in. This happens prior to the static analysis on the tarball, or if one doesn't exist yet. It doesn't and won't go into the spec file at all.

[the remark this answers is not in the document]
matthewrsj: Ah, right. I read this line before getting to the PV stuff maybe.

autospec/autospec.py
```diff
@@ -193,15 +195,29 @@ def main():
     parser.add_argument("--non_interactive", action="store_true",
                         default=False,
                         help="Disable interactive mode for package verification")
+
```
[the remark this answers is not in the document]
Gabib32: done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+                v = v.replace("${ROS_SPN}", bb_dict.get("ROS_SPN"))
+            elif "NAME" in bb_dict:
+                v = v.replace("${ROS_SPN}", bb_dict.get('NAME'))
+
+        bb_dict[k] = v
+
+
+def get_src_url(bb_dict):
+    if bb_dict.get('SRC_URI'):
+        bb_dict["URL"] = bb_dict.get("SRC_URI").split('\\')[0].split(';')[0]
```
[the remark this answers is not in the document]
Gabib32: Some SRC_URI variables have a value that is the source url and multiple patch files. A common .bb looks like this: 
SRC_URI = "https://github.com/ros/${ROS_SPN}/archive/${PV}.tar.gz;downloadfilename=${ROS_SP}.tar.gz \
           file://0001-Replaced-references-to-deprecated-Boost.TR1.patch \
           "
So the split(';') removes that 'downloadfilename..' part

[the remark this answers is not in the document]
matthewrsj: Ah so you first split on `'\\'` so you are only attempting the `split(';')` on the first line? That way you won't get any additional line junk in the URI if if the first line doesn't have a `';'`. Got it.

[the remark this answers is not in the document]
Gabib32: Yeah, exactly!

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+        value = value[1:-1]
+
+    # remove line continuations
+    value = "".join(value.split("\\"))
+    return value
+
+
+def update_inherit(line, bb_dict):
+    if 'inherits' in bb_dict:
+        bb_dict['inherits'].append(' '.join(line.split(' ', 1)[1:]))
```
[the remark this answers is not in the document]
Gabib32: inherits can be multiple lines, that always start with "inherit", but can have multiple values following it. For example:
inherit autotools update-alternatives
inherit autotools-brokensep
I was splitting them to remove the inherit, and then store as a list, with each line its own value. I assume order matters, so it would be bb_dict["inherit"] = ['autotools' 'update-alternatives', 'autotools-brokensep']. I'll do some research to see if they can each be stored as their own value in a list (splitting autotools and update-alternatives to be 2 values, not 1)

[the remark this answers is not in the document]
Gabib32: Inherits can have multiple lines, and multiple values on each line: 
inherit autotools update-alternatives
inherit autotools-brokensep

Currently its storing a list with ["autotools update-alternatives", "autotools-brokensep"] as opposed to each inherit being its own item. My reasoning for updating is so that the value doesn't get over written if there is more than one inherit.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def scrape_version(bb_dict):
+    # get version from latest filename added to dict, if exists
+    try:
+        bb_dict['VERSION'] = bb_dict.get('filename')[-1].split('_', 1,)[1].rsplit('.', 1)[0]
+    except IndexError:
+        print('file has no version: {}')
+
+    return bb_dict
```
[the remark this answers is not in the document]
Gabib32: done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def replace_pv(bb_dict):
+    for k, v in bb_dict.items():
+        if "${PV}" in v and "VERSION" in bb_dict:
```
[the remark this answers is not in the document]
Gabib32: Done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def get_src_url(bb_dict):
+    if bb_dict.get('SRC_URI'):
+        bb_dict["URL"] = bb_dict.get("SRC_URI").split('\\')[0].split(';')[0]
+
+
+def clean_values(value):
+    # remove beginning and end quote
+    if value.startswith('"') and value.endswith('"'):
```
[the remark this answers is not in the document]
Gabib32: Done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+        value = value[1:-1]
+
+    # remove line continuations
+    value = "".join(value.split("\\"))
```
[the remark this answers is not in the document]
Gabib32: Done

autospec/infile_bb_parser.py
Gabib32: Done

autospec/infile_handler.py
Gabib32: It is - I wasn't sure what the appropriate response should be if the infile is invalid. Do we want the autospec run to continue, or print error_fatal and exit?

matthewrsj: We should continue, by all means, but just `Error downloading url` is a pretty cryptic error message. The user should know that the infile could not be parsed and autospec is continuing its normal operations.

autospec/infile_bb_parser.py
Gabib32: Done

autospec/infile_bb_parser.py
Gabib32: I thought about this, but do you think the extra storing and double iteration would be costly?

matthewrsj: Not really, the bitbake recipes I've seen aren't that big. Besides, autospec already has to scan enormous build logs potentially several times per build so this would be a drop in the bucket. I'd rather have it easier to read and understand.

Gabib32: Do you think the storing of the list, and extra iteration would be costly?

matthewrsj: err, see above.

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+
+def bb_scraper(bb_fp, bb_dict):
+    todo = []
+
+    if bb_dict.get('filename'):
+        bb_dict = scrape_version(bb_dict)
+
+    for line in bb_fp:
+        cont = None
+        if line.strip() and not line.strip().startswith('#'):
```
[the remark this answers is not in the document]
Gabib32: done

autospec/infile_bb_parser.py
Gabib32: Done

autospec/infile_bb_parser.py
Gabib32: Done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+                    if line:
+                        cmd.append("# " + line)
+
+                    if depth == 0:
+                        break
+                    else:
+                        line = next(bb_fp).strip()
+
+                if cmd[1].startswith("# " + cmd_name) and cmd[-1] == '# }':
+                    cmd = cmd[2:-1]
```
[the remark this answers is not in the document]
Gabib32: Done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,202 @@
...
+                    depth = read_in_command(line, depth)
+                    if line:
+                        cmd.append("# " + line)
+
+                    if depth == 0:
+                        break
+                    else:
+                        line = next(bb_fp).strip()
+
+                if cmd[1].startswith("# " + cmd_name) and cmd[-1] == '# }':
```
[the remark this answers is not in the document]
Gabib32: Done

autospec/infile_bb_parser.py
Gabib32: done

autospec/infile_bb_parser.py
Gabib32: Yeah it was - removed.

autospec/infile_handler.py
Gabib32: +1 Done

autospec/infile_handler.py
Gabib32: Added github parsing
Done :)

autospec/infile_handler.py
Gabib32: Done

autospec/infile_handler.py
Gabib32: Done

autospec/infile_handler.py
Gabib32: Donee

autospec/infile_update_spec.py
Gabib32: Do we want to keep default_sum to the "No detailed summary available"?

Gabib32: Done

autospec/infile_update_spec.py
Gabib32: Done

autospec/infile_handler.py
Gabib32: Done

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,234 @@
...
+                    lines.append(line)
+                    break
+
+            line = " ".join(lines)
+
+        if line.startswith('inherit'):
+            bb_dict['inherits'].append(line.split(maxsplit=1)[1])
+            continue
+
+        match = pattern_match_line(line)
```
matthewrsj: You are compiling the same regular expression for every line in the file. You can compile that regular expression before this loop and just run search on the line (much faster). Maybe `pattern_match_line()` could return the compiled pattern instead of the result of the regex search?

autospec/infile_bb_parser.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def evaluate_expr(op, cur_val, assignment):
+    if not cur_val:
+        return assignment
+    return operators_dict.get(op).format(cur_val, assignment)
+
+
+def write_to_dict(bb_dict, m):
+
```
matthewrsj: Docstrings are missing on several important functions. I know much of autospec doesn't follow this standard but our new functions should.

autospec/infile_handler.py
```diff
@@ -0,0 +1,138 @@
...
+    This function determines whether the input is a file or a url. If it is a
+    url it checks that it is in the correct format (plaintext), downloads the
+    url to a temporary file and passes the file handler to be scraped.
+    If the input is a file, then it opens the file and passes the handler to
+    be scraped.
+
+    The type of parsing bitbake, inc, deb, etc is based on the file extension.
+    """
+
+    parse_type = parse_ext(indata)
```
matthewrsj: Parsing can fail at this point if the extension is not recognized. You should fail out of the `file_handler` here (by checking `parse_type`) instead of letting it fail farther along.

autospec/infile_handler.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if output_dict.get('filename'):
+        output_dict['filename'].append(indata)
+    else:
+        output_dict['filename'] = [indata]
+
+    if not os.path.isfile(indata):
+        # check that input is plain or raw text and not html
+        indata = check_url_content(indata)
+        indata = None
```
matthewrsj: Wait, `indata` is always `None` now?

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def update_licenses(bb_dict, specfile):
+    """
+    The specfile contains a list of licenses for a package, if the bitbake
+    license is not included in that list, add it.
+    """
+    if "LICENSE" in bb_dict:
+        if bb_dict.get("LICENSE").lower() not in [l.lower() for l in specfile.licenses]:
+            specfile.licenses.append(bb_dict.get("LICENSE"))
```
matthewrsj: `specfile.licenses` should be converted to a `set` in order to make adding licenses simpler, but that doesn't need to be done in this PR.

autospec/infile_update_spec.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    deps = set()
+    if bb_dict.get('DEPENDS'):
+        for dep in bb_dict.get('DEPENDS').split():
+            dep = re.match(r"(\$\{PYTHON_PN\}\-)?([a-zA-Z0-9\-]+)", dep).group(2)
+            if dep.endswith('-native'):
+                dep = dep[:-7]
+            deps.add(dep)
+
+    specfile.buildreqs = specfile.buildreqs.union(deps)
```
matthewrsj: how about just doing `specfile.buildreqs.add(dep)` for each one found instead of building a new set?

autospec/infile_handler.py
Gabib32: ooo good catch this was a really bad debugging thing I left in.

autospec/infile_update_spec.py
Gabib32: Noted

autospec/infile_update_spec.py
```diff
@@ -80,6 +87,7 @@ def write_cmd_files(bb_dict, dirname):
             with open(filename, 'a') as cmdfp:
                 cmdfp.write("# Infile parser added the following lines:\n")
                 cmdfp.write("\n".join(cmd) + "\n")
+            print_infile("File: {} changed".format(cmd_mappings.get(cmd_name)))
```
matthewrsj: Could this say something like `"Suggestions added to {} file"` to make it a little more intuitive for the user?

matthewrsj:
+1, good work.

I'm opening #76 and assigning it to @Gabib32 to track adding a section to the README.

PR #69, opened 2017-11-17
Add autoupdate flag to options.conf

matthewrsj:
This flag is a no-op for autospec itself, but allows humans to indicate
that a package is trusted enough to automatically release new updates of
the package. In practice this should be used by a tool that runs
autospec in an automated manner to push the package to a build server.


bryteise:
+1

fenrus75:
+1

PR #70, opened 2017-11-22
cmake: add cmake_srcdir configuration

dorileo:
There are some packages that will ship multiple packages in the same
tarball. This configuration works around that and run cmake on the
correct source sub directory.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -42,6 +42,7 @@ urlban = ""
 extra_make = ""
 extra_make_install = ""
 extra_cmake = ""
+cmake_srcdir = ""
 prep_append = []
 subdir = ""
 install_macro = "%make_install"
@@ -481,6 +482,7 @@ def parse_config_files(path, bump, filemanager):
     global extra_make
     global extra_make_install
     global extra_cmake
+    global cmake_srcdir
     global prep_append
     global subdir
     global install_macro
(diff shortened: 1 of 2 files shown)


dorileo:
@matthewrsj just to give you some more context, this package has the issue I mention:

https://github.com/ros/ros

dorileo:
@matthewrsj 

> You need to actually define a  `cmake_srcdir` member variable for the specfile class.

[Done]

> You also need to make sure the default .. is used when the configuration file is not supplied.

[Done] It was there already, but now I moved to the `cmake_srcdir` member variable initialization.


Thanks...

mrkz:
@dorileo  would this option be enabled on the options.conf of the package?

If so, @matthewrsj, @dorileo   Is the option written/documented correctly when the options.conf is written? (haven't got up to date with the internals of this part of the code)

matthewrsj:
@mrkz no this would be an additional text configuration file called `cmake_srcdir` that exists in the package directory. Unfortunately many of these configuration files are documented internally but have not been documented in the README.

In fact Autospec could probably use some revamped documentation that exists outside of the README if someone has time to implement that. I've opened #72 to track.

autospec/config.py
```diff
@@ -725,6 +731,7 @@ def load_specfile(specfile):
     specfile.extra_make = extra_make
     specfile.extra_make_install = extra_make_install
     specfile.extra_cmake = extra_cmake
+    specfile.cmake_srcdir = cmake_srcdir
```
matthewrsj: What happens if the `cmake_srcdir` file does not exist so `specfile.cmake_srcdir` gets the current contents of the `cmake_srcdir` variable. Won't the `cmake_srcdir` variable have an empty string and overwrite the default `..`?

Maybe you could check before assignment:
```
specfile.cmake_srcdir = cmake_srcdir or specfile.cmake_srcdir
```
to make sure you aren't overwriting the default with an empty string.

dorileo: sure

dorileo:
> Maybe you could check before assignment:
> specfile.cmake_srcdir = cmake_srcdir or specfile.cmake_srcdir

[Done]

PR #71, opened 2017-11-23
failed_commands: add pattern to detect ldap.h

mrkz:
with this change autospec detects when ldap.h is missing
to add the openldap-dev build requirement.


matthewrsj:
+1

PR #73, opened 2017-11-28
Print commitmessage as a string

matthewrsj:
commitmessage is a list and is being printed as a string representation
of a list. Join it to a string to the print looks sane.


phmccarty:
+1

matthewrsj:
woohoo

PR #82, opened 2017-12-05
Import new public key (Thomas Orgis)

phmccarty:
This key is needed to verify the signature for 'mpg123' release tarballs.


fenrus75:
We do need to add a feature to allow the key to be in package git, that
would allow us to only need to centralize common project keys

matthewrsj:
@fenrus75 I agree. https://github.com/clearlinux/autospec/issues/83 opened to track.

matthewrsj:
+1

PR #85, opened 2017-12-14
Remove verbose flag for make line

matthewrsj:
Autospec does not parse the verbose output from make, so V=1 is not
necessary and slows down autospec build log parsing considerably. If
users want to debug a build at that level they should add the V=1 back
manually.


matthewrsj:
This approach was discussed offline and approved. Merging.

PR #87, opened 2017-12-19
Change default python buildpattern to distutils3

matthewrsj:
Change default python buildpattern from distutils23 (both python2 and
python3) to distutils3 (python3 only). This is a step toward deprecating
python2 usage in distro packages. Any packages that rely on python2
should have "distutils23" added to the build_pattern configuration file
to maintain their python2 subpackage (or, preferrably, built with
python3 only as is default).


phmccarty:
+1

PR #88, opened 2017-12-19
more catkin patterns

dorileo:
Depending on the dependencies name catkin will break the line or not,
this new patterns will catch more errors.

Signed-off-by: Leandro Dorileo <leandro.maciel.dorileo@intel.com>


matthewrsj:
LGTM

PR #96, opened 2018-01-11
Remove no-longer-used functional test suite

matthewrsj:
This test suite is extremely hard to keep up-to-date and is no longer as
valuable now that unit tests have been written for much of the code
base. Actual functional testing can be done against package repos
without requiring them to be included in the autospec repository.


bryteise:
+1 I look forward to getting the testing moved to a more practical place.

PR #97, opened 2018-01-12
Turn off ConfigParser string interpolation

matthewrsj:
configparser.ConfigParser attempts to interpolate "%" characters in the
values of the key-value pairs when parsing. This is absolutely unwanted
behavior, especially when parsing URLs such as

<sourceforge url>/joe-editor/files/JOE%20sources/<joe tar>

which cause an uncaught exception.

Turn off interpolation by passing the interpolation=None argument to
ConfigParser.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,9 @@ test_util:
 test_infile_parser:
 	PYTHONPATH=${CURDIR}/autospec python3 -m unittest discover -b -s tests -p 'test_infile_*'
 
+test_general:
+	PYTHONPATH=${CURDIR}/autospec python3 tests/test_general.py
+
 unittests:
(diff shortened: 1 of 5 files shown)


bryteise:
Seems fine, unit tests to keep this from regressing?

matthewrsj:
wouldn't that just be testing ConfigParser? The only thing changed here is adding an argument to the ConfigParser() call, so a unit tests to test that would just be testing that ConfigParser is doing what it is documented to do.

bryteise:
Maybe just script that looks for any use of ConfigParser that isn't passing the disable option?

matthewrsj:
ok, that seems fine.

matthewrsj:
@bryteise test added.

matthewrsj:
Had off-github confirmation of approval from @bryteise

PR #98, opened 2018-01-12
Check for ConfigParser regressions in all source files

matthewrsj:
Was only checking one file in last commit. This checks all files under autospec/


matthewrsj:
Merging - simple test fix/improvement.

PR #103, opened 2018-01-13
Fixups for rust/cargo building

bryteise:
Improve rust/cargo support by:
1) Better handling only toplevel Cargo configuration files for dependency
lookup.
2) Adding runtime requirements for buildreq packages in order to get
dependencies from non directly dependent packages resolved.
3) Add cargo as a buildreq for rust packages to handle the rustc/cargo
split.
4) Add cargo detection from failed_commands.
5) Fix rust source directory destination to not append version in case
the package name already had version information in it.
6) Get name and version information from crates.io package files.

---
Can split up this into individual patches if wanted but thought
generic rust support would be easier to see in one patch.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -224,6 +224,7 @@ def parse_cargo_toml(filename):
     """
     global cargo_bin
     buildpattern.set_build_pattern("cargo", 1)
+    add_buildreq("cargo")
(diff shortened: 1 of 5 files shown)


matthewrsj:
Without being very familiar with rust and cargo, these changes are innocuous enough and make sense to me. +1

PR #105, opened 2018-01-17
Add 32bit versions of make/make_install args

bryteise:
Add make32_args and make32_install_args config file handling to allow
specific options to be passed that are specific to the 32bit builds.


autospec/config.py
```diff
@@ -484,7 +486,9 @@ def parse_config_files(path, bump, filemanager):
     global profile_payload
     global config_opts
     global extra_make
+    global extra32_make
```
matthewrsj: another `global...` 😭 

JK, I'm the biggest offender here.

PR #106, opened 2018-01-17
Respect [autospec][urlban] for archive URLs

### LINKED ISSUE #104: urlban not recognized for --archives values
If I set [autospec][urlban], I would expect the urlban subsitution to happen for the URL(s) passed to --archives.

matthewrsj:
Fixes #104


matthewrsj:
Nope, there is an error here.

Issue #83, opened 2017-12-05
Make default keyring the target repository


matthewrsj:
Tell GPG to check the package repository (target) first before using the autospec keyring.

matthewrsj:
Fixed by #109

PR #109, opened 2018-01-18
Change default keyring to package directory

matthewrsj:
Instead of relying on a centralized keyring that resides in the autospec
repository, make the default keyring the package directory itself. Any
key imports will be to the package repository and will be committed as
part of the automatic autospec commit.

Retain the autospec repo keyring as a fallback keyring when the public
key does not exist in the package repo.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -45,6 +45,7 @@ RUBYORG_API = "https://rubygems.org/api/v1/versions/{}.json"
 PYPIORG_API = "https://pypi.python.org/pypi/{}/json"
 KEYID_TRY = ""
 KEYID = ""
+IMPORTED = ""
 EMAIL = ""
 GNUPGCONF = """keyserver keys.gnupg.net"""
 CMD_TIMEOUT = 20
@@ -606,6 +607,7 @@ class InputGetter(object):
 
 
 def attempt_key_import(keyid, key_fullpath):
+    global IMPORTED
     print(SEPT)
     ig = InputGetter('\nDo you want to attempt to import keyid {}: (y/N) '.format(keyid))
     import_key_answer = ig.get_answer()
@@ -630,6 +632,7 @@ def attempt_key_import(keyid, key_fullpath):
         print("\n", content)
         ig = InputGetter(message='\nDo you want to keep this key: (Y/n) ', default='y')
         if ig.get_answer() is True:
+            IMPORTED = content
             return True
         else:
             os.unlink(key_fullpath)
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -47,7 +47,6 @@ KEYID_TRY = ""
 KEYID = ""
 EMAIL = ""
 GNUPGCONF = """keyserver keys.gnupg.net"""
-PUBKEY_PATH = '/'.join([os.path.dirname(os.path.abspath(__file__)), "keyring", "{}.pkey"])
 CMD_TIMEOUT = 20
 ENV = os.environ
 INPUT_GETTER_TIMEOUT = 60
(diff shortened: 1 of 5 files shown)


matthewrsj:
Fixes #83

autospec/pkg_integrity.py
```diff
@@ -754,6 +759,8 @@ def get_integrity_file(package_path):
 
 
 def check(url, download_path, interactive=True):
+    global PUBKEY_PATH
+    PUBKEY_PATH = os.path.join(download_path, "{}.pkey")
```
bryteise: This feels not right. I'd think that the GPGVerifier should probably store the PUBKEY_PATH in the object instead of having it set in check and be global (and also having the string contain an embedded format specifier).

from_disk and from_url are the only callers who need to create the object and both have the path in their argument list. Also reduces a global, what do you think?

matthewrsj: Agree, done

matthewrsj:
> Not related but should we start to clean up old public keys and old tar file signatures that are no longer used?

Yes I think so, I was going to remove my public key after this is merged as a separate PR.

autospec/pkg_integrity.py
```diff
@@ -468,15 +471,20 @@ def verify(self, recursion=False):
...
+        pub_key = self.get_pub_key(self.pubkey_path)
         EMAIL = parse_key(pub_key, r':user ID packet: ".* <(.+?)>"\n')
-        if not pub_key or os.path.exists(pub_key) is False:
-            key_id = get_keyid(self.package_sign_path)
-            self.print_result(False, 'Public key {} not found in keyring'.format(key_id))
-            if self.interactive is True and recursion is False and attempt_key_import(key_id):
-                print(SEPT)
-                return self.verify(recursion=True)
-            return None
+        if not pub_key or not os.path.exists(pub_key):
```
bryteise: Hrm still seems clunky... Also I don't think EMAIL should be set until after this block since it may not be valid (and since EMAIL is global could potentially be a problem later on).

What about having the pub_key and fallback_pubkey be set in the object, don't even need the get_pub_key function at all then. Gets rid of the extra format for key_fullpath too I think. You can get the keyid as part of the init  since that wouldn't ever change.

autospec/pkg_integrity.py
```diff
@@ -468,16 +471,19 @@ def verify(self, recursion=False):
...
-        if not pub_key or os.path.exists(pub_key) is False:
-            key_id = get_keyid(self.package_sign_path)
-            self.print_result(False, 'Public key {} not found in keyring'.format(key_id))
-            if self.interactive is True and recursion is False and attempt_key_import(key_id):
-                print(SEPT)
-                return self.verify(recursion=True)
-            return None
-        sign_status = verify_cli(pub_key, self.package_path, self.package_sign_path)
+        pubkey_loc = self.pubkey_path
+        EMAIL = parse_key(pubkey_loc, r':user ID packet: ".* <(.+?)>"\n')
```
bryteise: EMAIL should be moved to just before sign_status is assigned I think.

autospec/pkg_integrity.py
```diff
@@ -468,16 +471,19 @@ def verify(self, recursion=False):
...
+        if not os.path.exists(pubkey_loc):
+            print_info("public key not in default keyring, checking fallback keyring")
+            pubkey_loc = self.fallback_pubkey_path
+            EMAIL = parse_key(pubkey_loc, r':user ID packet: ".* <(.+?)>"\n')
```
bryteise: That way you don't need to have this line I'm pretty sure.

bryteise:
Looks good in theory now aside from my last comment, just need to get tests passing again.

matthewrsj:
@bryteise sorry about that. Fixed now.

bryteise:
+1

PR #110, opened 2018-01-18
Add --fingerprint output when importing pubkeys

### LINKED ISSUE #79: Print fingerprint when importing new public keys
When autospec tries to verify a signature created with an unrecognized key pair, it will download the public key and display some of its metadata so the user can decide whether or not to import the key. It would be nice to also print the key's fingerprint so that the user can cross-reference a PGP key server and make a more informed decision on this process.

matthewrsj:
Fixes #79

Print the key's fingerprint so the user can cross-reference a PGP key
server and make a more informed decision to import the public key.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -130,10 +130,20 @@ class GPGCli(object):
 
     def display_keyinfo(self, keyfile):
         args = self.args + ['--list-packet', keyfile]
-        output, err, code = self.exec_cmd(args)
-        if code == 0:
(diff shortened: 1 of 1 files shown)


matthewrsj:
That is a good point about 60 seconds not being long enough, but I think that needs some discussion. Do we want to wait indefinitely so the user can research? Do we want to just exit right then and provide the user with the command to import themselves? Or do we just want to give them some amount of time to make a decision (like 60 seconds...?)

Regardless, out of scope for this PR. Merging...

PR #111, opened 2018-01-18
Remove Matthew Johnson public key

matthewrsj:
Now that these keys can exist in the individual package repos, remove my
key from the global autospec keyring.


matthewrsj:
Merging this...

PR #112, opened 2018-01-19
Improve instructions for --skip-git

matthewrsj:
The user should be told they need to add relevant files before they
commit.


bryteise:
+1 do we eventually want to add which files are relevent/required for autospec?

matthewrsj:
Yes a `git_add(noop=False)` function might be nice. If `noop == True` it could just print out what files to add. Regardless, a feature for another time.

PR #122, opened 2018-02-14
Add support for qmake-based projects

thiagomacieira:
qmake uses project files called *.pro, which we detect at the top-level,
so long as there's no configure script also present. We scan the *.pro
files for QT{,_PRIVATE,_FOR_CONFIG} += lines, which will give us the
extra dependencies, then convert from the qmake internal name to the
pkgconfig name.

This build pattern works for all Qt 5.10 modules, except that it
concludes that a module depends on itself, if it has a library depending
on another library that it's also building (example: libQtQuick.so
depends on libQtQml.so). For those modules, it's necessary to list the
libraries that are product of the package in the pkgconfig_ban file.


autospec/qt-modules.txt
```diff
@@ -0,0 +1,61 @@
+3danimation Qt53DAnimation
```
matthewrsj: Can you please change the format of this file to
```
# comment explaining the purpose and format of this file
key, value
key, value
...
```
to be consistent with the other configuration files in autospec (`failed_commands`, `maven_jars`). Then the file can be read in using `config.read_pattern_conf()`.

matthewrsj: You can also drop the `.txt` extension as it isn't needed.

thiagomacieira: Will do.

matthewrsj:
Thanks for making those changes, +1.

PR #124, opened 2018-02-15
Remove the shared keyring.

busykai:
Keys now must be stored in the individual packages. autospec will import them
there (as opposed to this central location). Packages were updated with their
current pub keys where possible.

@matthewrsj


busykai:
Heh, will fix the test now.

busykai:
is a fallback that is not shared automatically (like keyring we distributed with autospec) useful at all? that is, if i have my local fallback location, it makes autospec'ing packaging non-reproducible on other machines.

i actually thought about removing the code that looks into autospec's keyring altogether (after a grace period).

matthewrsj:
I'd actually be completely fine with removing the fallback keyring logic. No need for a grace period, you already removed all the fallback keys. It just should be noted that right now we provide a fallback in the code but not in reality.

I think it makes sense to just allow the keys to sit in the package trees. These pubkeys should be approved on a per-package basis not on a per-key basis.

PR #126, opened 2018-02-21
files: exclude /usr/lib{,32,64}/pkgconfig files

mrkz:
These files are not required (as they're just copies of
/usr/lib64/pkgconfig/ files). So we exclude them by default.

### DIFF
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -178,6 +178,11 @@ class FileManager(object):
             self.push_package_file(newfn, "attr")
             self.excludes.append(filename)
 
+        # exclude all lib{,32,64}/haswell pkgconfig files
+        hsw_pkgconfig = re.compile(r"/usr/lib(|32|64)/haswell/pkgconfig/.+")
+        if hsw_pkgconfig.search(filename):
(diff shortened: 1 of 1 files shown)


autospec/files.py
```diff
@@ -178,6 +178,11 @@ def push_file(self, filename):
             self.push_package_file(newfn, "attr")
             self.excludes.append(filename)
 
+        # exclude all lib{,32,64}/haswell pkgconfig files
+        hsw_pkgconfig = re.compile(r"/usr/lib(|32|64)/haswell/pkgconfig/.+")
```
bryteise: Could we instead exclude all pkgconfig files not in /usr/lib64/pkgconfig rather than need to keep poking this every time we add another instruction set.

mrkz: You mean something like replacing
```
r"/usr/lib(|32|64)/haswell/pkgconfig/.+"
```
with
```
r"/usr/lib(|32|64)/[a-zA-Z0-9]+/pkgconfig/.+"
```
?

mrkz:
No longer required (those file are now removed by clr-rpm-config)

PR #127, opened 2018-02-21
A number of autospec improvements

bryteise:
Taken from a patchset from @fenrus75

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -223,19 +223,21 @@ def head_request(url):
 
 
 def get_signature_url(package_url):
-    if '://pypi.' in package_url[:13]:
+    if 'samba.org' in package_url:
+        return package_url + '.asc'
+    elif '://pypi.' in package_url[:13]:
         return package_url + '.asc'
     elif 'mirrors.kernel.org' in package_url:
         return package_url + '.sig'
     else:
         try:
-            st = head_request(package_url + '.sig')
-            if st == 200 or st == 302:
-                return package_url + '.sig'
             st = head_request(package_url + '.asc')
             if st == 200 or st == 302:
                 return package_url + '.asc'
-            st = head_request(package_url + '.asc')
+            st = head_request(package_url + '.sig')
+            if st == 200 or st == 302:
+                return package_url + '.sig'
+            st = head_request(package_url + '.sign')
             if st == 200 or st == 302:
                 return package_url + '.sign'
         except:
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -1275,8 +1275,33 @@ qmake {}QMAKE_CFLAGS=\"$CFLAGS\" QMAKE_CXXFLAGS=\"$CXXFLAGS\" QMAKE_LFLAGS=\"$LD
         self.write_variables()
         self._write_strip('CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS" LDFLAGS="$LDFLAGS" meson --prefix /usr --buildtype=plain {0} builddir'.format(config.extra_configure))
         self._write_strip("ninja -v -C builddir")
+
+        if config.config_opts['32bit']:
+            self._write_strip("pushd ../build32")
+            self._write_strip('export PKG_CONFIG_PATH="/usr/lib32/pkgconfig"')
+            self._write_strip('export CFLAGS="$CFLAGS -m32"')
+            self._write_strip('export CXXFLAGS="$CXXFLAGS -m32"')
+            self._write_strip('export LDFLAGS="$LDFLAGS -m32"')
+            self._write_strip('CFLAGS="$CFLAGS -m32" CXXFLAGS="$CXXFLAGS -m32" LDFLAGS="$LDFLAGS "'
+                              '"-m32 PKG_CONFIG_PATH="/usr/lib32/pkgconfig" meson "'
+                              '"--libdir=/usr/lib32 --prefix /usr --buildtype=plain {0} builddir'
+                              .format(config.extra_configure))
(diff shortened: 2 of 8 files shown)


autospec/pkg_integrity.py
```diff
@@ -223,19 +223,21 @@ def head_request(url):
...
 def get_signature_url(package_url):
-    if '://pypi.' in package_url[:13]:
+    if 'samba.org' in package_url:
+        return package_url + '.asc'
+    elif '://pypi.' in package_url[:13]:
         return package_url + '.asc'
     elif 'mirrors.kernel.org' in package_url:
         return package_url + '.sig'
     else:
         try:
```
busykai: This bothers me so much, maybe it's the appropriate moment to re-write the whole `try:` body as:
```python
try:
    iter = (package_url + "." + ext for ext in ("sig", "asc", "sign"))
        for sign_url in iter:
            if head_request(sign_url) in (200, 302):
                return sign_url
except:
    pass
```

matthewrsj:
Commit 1: I'd rather see this consistently implemented throughout the code base before applying this. I don't think applying this selectively helps with readability.

matthewrsj:
Commit 2: +1

matthewrsj:
Commits 3-6: +1

matthewrsj:
Commit 7: -1, look just a few lines lower. This is implemented already.

autospec/pkg_integrity.py
matthewrsj: Agree that is much better, but maybe not as part of this patchset. Feel free to open a PR for that change though.

matthewrsj:
Commits 8 and 11: +1

Commits 9 and 10 require further review.

autospec/pkg_integrity.py
busykai: Opened #128. It can wait till this merges so @bryteise wouldn't have to resolve.

bryteise:
@matthewrsj For commit 1, do you want the types literally everywhere or just for global values before accepting? I'm dropping the patch from this series though in the mean time.

bryteise:
@matthewrsj Once 9 and 10 are reviewed I'll push and updated set with 1 and 7 removed.

matthewrsj:
I guess just for all global values would be a good start, though I'm not entirely convinced these are necessary at all. From what I can tell it is just for readability and linting warnings, no? Anyways that can be in a separate PR to discuss there.

> we need to start the typing somewhere tbh

Ok, but let's at least cover all of *something* not just some of the globals. At least cover all the globals before merging this.

bryteise:
@fenrus75 I do note that the makes autospec python 3.6+ which we've stepped back from before.

matthewrsj:
> @fenrus75 I do note that the makes autospec python 3.6+ which we've stepped back from before.

Good point, we deprecated the new format strings (`f"..."`) to support at least 3.5

autospec/specfiles.py
```diff
@@ -431,8 +436,11 @@ def write_variables(self):
                 flags.extend(["-Os", "-ffunction-sections", "-fdata-sections", "-fno-semantic-interposition"])
         if config.config_opts['security_sensitive']:
             flags.append("-fstack-protector-strong")
+            flags.append("-mzero-caller-saved-regs")
```
matthewrsj: This change is not related to the AVX512 code this patch applies.

autospec/specfiles.py
```diff
@@ -581,6 +594,14 @@ def write_cmake_install(self):
             self._write_strip("popd")
             self._write_strip("rm -f %{buildroot}/usr/bin/*")
 
+        if config.config_opts['use_avx512']:
+            self._write_strip("mkdir -p %{buildroot}/usr/lib64/haswell/avx512_1")
+            self._write_strip("pushd clr-build-avx512")
+            self._write_strip("%s %s || :\n" % (self.install_macro, self.extra_make_install))
+            self._write_strip("mv %{buildroot}/usr/lib64/*so* %{buildroot}/usr/lib64/haswell/avx512_1 || :")
+            self._write_strip("popd")
```
matthewrsj: optional, can this be moved one line up?

autospec/specfiles.py
```diff
@@ -838,6 +885,35 @@ def write_autogen_pattern(self):
                                       config.extra_configure32))
             self.write_make_line(True)
             self._write_strip("popd")
+
+        if config.config_opts['use_avx2']:
```
matthewrsj: Not really part of AVX512, but fine because it is the same type of change.

autospec/specfiles.py
```diff
@@ -1169,8 +1275,32 @@ def write_meson_pattern(self):
         self.write_variables()
         self._write_strip('CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS" LDFLAGS="$LDFLAGS" meson --prefix /usr --buildtype=plain {0} builddir'.format(config.extra_configure))
         self._write_strip("ninja -v -C builddir")
+
+        if config.config_opts['32bit']:
```
matthewrsj: Here and below, different change from AVX512 changes.

autospec/specfiles.py
bryteise: You want the popd moved up one line? Do you want that moved up in the avx2 code above this as well?

matthewrsj:
Commit 10: Eh, sure. I'm not sure I like the fallback being the homedir but I don't really have a better alternative, so +1 for that.

autospec/specfiles.py
matthewrsj: sure, optional like I said though. little deal.

bryteise: Skipping for now.

bryteise:
@matthewrsj Updated.

PR #128, opened 2018-02-22
Simplify the logic of signature url lookup.

busykai:
Besides the simplification, it fixes a functional bug: ".sign" extension was
never checked properly.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -231,15 +231,10 @@ def get_signature_url(package_url):
         return package_url + '.sig'
     else:
         try:
-            st = head_request(package_url + '.asc')
-            if st == 200 or st == 302:
-                return package_url + '.asc'
-            st = head_request(package_url + '.sig')
-            if st == 200 or st == 302:
-                return package_url + '.sig'
-            st = head_request(package_url + '.sign')
-            if st == 200 or st == 302:
-                return package_url + '.sign'
+            iter = (package_url + "." + ext for ext in ("asc", "sig", "sign"))
+            for sign_url in iter:
+                if head_request(sign_url) in (200, 302):
+                    return sign_url
         except:
             pass
     return None


fenrus75:
this still does not work well fwiw

the whole "302" is a mistake and caused the original changes in the other patch.
There are many sites that always 302 to some other place independent of the file existing or not.

we must really make this stuff follow redirects and then only look at 200

matthewrsj:
@fenrus75 true, simple enough to `c.setopt(c.FOLLOWLOCATION, True)`

imo this can be done as part of this PR or another one.

busykai:
@fenrus75, @matthewrsj 302 no, FOLLOWLOCATION will not work with HEAD. Just as 302 may not mean that the file will be there, some servers (most notably Amazon) will respond with an error on HEAD, but GET will pass. 

maybe we should get rid of using HEAD instead.

busykai:
@fenrus75, I'd be interested to look into specific example where 302 on HEAD results in subsequent failure of GET. could you give me one?

matthewrsj:
It might just make sense to download the first hit, so no head request required. These are ordered in preference order as well are they not?

busykai:
> It might just make sense to download the first hit, so no head request required.

Agree. That's what I meant by using `GET` as opposed to `HEAD`.

bryteise:
I'd be fine with just downloading, the HEAD request isn't saving much in this case.

busykai:
I'd be fine as well. It's not the purpose of this PR though. I do not think the change I'm proposing breaks anything, rather otherwise.

bryteise:
Yes, for this change I'm +1 though I'm guessing we end up reworking a number of things to get to where we want to be.

autospec/pkg_integrity.py
```diff
@@ -231,15 +231,10 @@ def get_signature_url(package_url):
...
-            st = head_request(package_url + '.asc')
-            if st == 200 or st == 302:
-                return package_url + '.asc'
-            st = head_request(package_url + '.sig')
-            if st == 200 or st == 302:
-                return package_url + '.sig'
-            st = head_request(package_url + '.sign')
-            if st == 200 or st == 302:
-                return package_url + '.sign'
+            iter = (package_url + "." + ext for ext in ("sig", "asc", "sign"))
```
bryteise: So I'm just noticing this now but the order changed to prefer asc over sig (I'm unsure how important that is mind you, maybe @fenrus75 can comment on that) and this commit is changing it back.

busykai: there's a unit test that would fail if `.sig` is not preferred.

busykai: ok, so it has been changed with the recent merge, so the tests do not care what's the preferences. i'll just make it the same as the current master.

busykai:
I'm hitting the `403 Forbidden` with github-hosted artifacts again. Apparently, #173 has overriden the behavior... So I'll just fix it once and for all...

phmccarty:
@busykai FWIW, I always see `403 Forbidden` when issuing HEAD requests for artifacts that are linked from Github but are actually hosted in an S3 bucket. I don't see this issue for artifacts that are hosted by Github.

busykai:
@phmccarty, S3 is always the case when you upload artifacts. see [clr-boot-manager](https://github.com/clearlinux/clr-boot-manager/releases) or [clrtrust](https://github.com/clearlinux/clrtrust/releases). Since the code in question takes care of signatures, there's no way (AFAIK) to avoid uploading signatures on github.

phmccarty:
@busykai Ah sorry, I didn't read through the comment history. Never mind me :-)

Doing a `GET` is definitely the right way to go for the signatures, so this looks like a regression.

PR #132, opened 2018-03-06
Add new option to customize Mock config

phmccarty:
Instead of hardcoding the "-r clear" option to specify the Mock config, make this configurable with a new -m/--mock-config command line option.

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -203,7 +203,7 @@ def get_mock_cmd():
     return 'sudo /usr/bin/mock'
 
 
-def package(filemanager, cleanup=False):
+def package(filemanager, mockconfig, cleanup=False):
     global round
     global uniqueext
     round = round + 1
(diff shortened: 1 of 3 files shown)


autospec/build.py
```diff
@@ -203,7 +203,7 @@ def get_mock_cmd():
     return 'sudo /usr/bin/mock'
 
 
-def package(filemanager, cleanup=False):
+def package(filemanager, cleanup=False, mockconfig=False):
```
matthewrsj: You set mockconfig to be the bool `false` but then treat it as a string below. Please set this to be `""` by default instead.

autospec/build.py
```diff
@@ -218,20 +218,25 @@ def package(filemanager, cleanup=False):
         uniqueext = tarball.name
         cleanup_flag = "--no-cleanup-after"
 
+    if mockconfig:
+        config_option = "-r {}".format(mockconfig)
+    else:
```
matthewrsj: When will this ever happen? If the user doesn't pass `-m` then `mockconfig` will be set to `"clear"`. I believe `-m` will require an argument with the way you have it set up so I'm not seeing how mockconfig could ever be empty (or `False`, the way you use different types is confusing.)

phmccarty: It could never be empty with the current state of the code, true. Would you propose that I don't set a default value for `mockconfig` then?

matthewrsj: no I just propose you always set
```python
config_option = "-r {}".format(mockconfig)
```

phmccarty: I left a few things unsaid in my comment, but this is what I implied in the end. Okay, will fix.

PR #133, opened 2018-03-09
Add support for [un]installation scriptlets

MarioCarrilloA:
These commits add support for pre/post/preun/postun scriptlets.
This in order to control the RPM package [un]installation.

The serie of commands can be established in a file according to the
functionality, this relation is described below:

| Scriptlet | File name      |
| --------- | -------------- |
| %pre      | pre_install    |
| %preun    | pre_uninstall  |
| %post     | post_install   |
| %postun   | post_uninstall |

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -251,6 +251,26 @@ install_macro
   The contents of this file be used instead of the automatically detected
   ``install`` routine, i.e. use this if ``%make_install`` is insufficient.
 
+pre_install
+  The contents of this file be executed before ``install`` routine. This file
+  specifies a serie of commands that can be executed in order to set a required
+  environment before the installation stage.
+
+post_install
+  The contents of this file be executed after ``install`` routine. This file
+  specifies a serie of commands that can be executed in order to remove unneeded
+  stuff in the environment after the installation stage.
+
+pre_uninstall
+  The contents of this file be executed before to remove the package. This file
+  specifies a serie of commands that can be executed in order to set a required
+  enviroment before the uninstallation stage.
+
+post_uninstall
+  The contents of this file be executed after to remove the package. This file
+  specifies a serie of commands that can be executed in order to remove unneeded
+  stuff in the environment after the uninstallation stage.
+
 subdir
   Not all packages have their ``Makefile``'s available in the root of the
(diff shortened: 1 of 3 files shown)


phmccarty:
On Clear Linux, there are virtually no packages that define scriptlets, and this is intentional. Is this feature intended to support non-Clear Linux use cases?

MarioCarrilloA:
Hi @phmccarty , yeah, you are right, stateless feature defines that behavior, however the objective of this PR is to use `autospec` for non-Clear systems. The scriptlets/commands will be added  inside the SPEC file just in the case the helper files (pre_install, post_install etc) exist, this is in order to not break the current stateless packages, what do you think?

fenrus75:
If we end up going this way, we should allow a site configuration (in
common/ whatever) that bans these and will not let them be picked up, so
that we can enforce our no-%post policy centrally for Clear Linux

fenrus75:
And it's not because of stateless. We have this policy because we do not
use RPMs to distribute software (we use swupd), only to build software.
%post usage would break many of the assumptions and invariants we have for
swupd....
swupd has it's own mechanisms to do things post install.

bryteise:
Ah yes I expected a stateless field in options.conf to be added as part of this change that would then use the files if statless was turned off.

MarioCarrilloA:
Thanks @bryteise , expect a new version :)

fenrus75:
I'd not want this in per package options.conf
we need a central off switch in the overall distro configuration scripts
that autospec reads.

%post is not allowed in CL and we need to avoid accidents, not just for our
users, but also for people who mixes
(so it needs to live in common/)

mrkz:
uh, probably not in `options.conf` but in `autospec.conf` instead

matthewrsj:
@MarioCarrilloA are you still working on a new version of this?

matthewrsj:
@MarioCarrilloA are you still working on this or should the PR be closed?

MarioCarrilloA:
Hi @matthewrsj, I am not working on this for the moment, could be better to close it, thanks

matthewrsj:
Ok, thanks

PR #140, opened 2018-03-26
Remove fallback keyring from GPG verifier.

busykai:
This code paths is not supposed to be used since the keyring has been
removed from the autospec itself altogether. This is a clean up.


busykai:
done. just removed the word "default" from it. the rest of the sentence still holds.

PR #141, opened 2018-03-27
Fixup meson 32 bit flags

bryteise:
Quoting for meson 32bit builds was incorrect in a few places, causing
environment variables to be split and syntax errors in the command
with double quotes.


busykai:
LGTM. (Now that I've unintentionally looked at it. ;)

bryteise:
Hah, thanks =).

PR #143, opened 2018-04-05
add meson support for configure{32,64} files

### LINKED ISSUE #142: support configure32 configure64 files in meson build pattern
currently, the meson pattern supports only the `configure` file, disabling the ability of setting 32bits only or 64bits only configuration options.

mrkz:
Previous to this commit, meson build only noticed about the
configure file, which is used for both 64/32 builds. However,
autospec can manage to have specific configurations for
64 bits build or 32 bits build.

This commit enhances the meson build pattern, enabling the support
of specifig 64/32 build flags.

This fixes #142


mrkz:
@matthewrsj any chance you can review/merge this?

PR #153, opened 2018-05-09
Update mzcaller-saved option

VictorRodriguez:
Latest update of GCC 8 has a change in the patch zero-regs-gcc8.patch,
now instead of having just the option -mzero-caller-saved-regs (which add
the fnctionality to emit zero touched caller-saved general registers
upon function return) it now adds -mzero-caller-saved-regs=[skip|used|all]

In clear linux we will use the option:

-mzero-caller-saved-regs=used

This option emit a zero only on used caller-saved general registers upon
function return.


matthewrsj:
+1, but I understand this needs to be coordinated with the GCC update?

VictorRodriguez:
@matthewrsj GCC 8 is up in CLR since a minute ago :)

PR #161, opened 2018-05-22
Add a .gitignore to package repos

phmccarty:
There are several useful patterns to add for package repos managed by autospec, so dump those in a .gitignore.

Note that the final list entry (empty string) forces a newline to be added to the .gitignore file.

### DIFF
--- a/autospec/git.py
+++ b/autospec/git.py
@@ -20,11 +20,12 @@
 #
 
 import subprocess
+import os
 
 import build
 import buildpattern
 import tarball
-from util import call
+from util import call, write_out
 import config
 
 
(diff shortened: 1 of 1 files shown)


autospec/git.py
```diff
@@ -95,6 +96,23 @@ def commit_to_git(path):
...
+        ".repo-index",
+        "*.log*",
+        "*.tar*",
+        "*.tgz",
+        "*.zip",
+        "commitmsg",
+        "results/",
+        ""
+    ]
+    write_out(os.path.join(path, '.gitignore'), '\n'.join(ignorelist))
```
matthewrsj: Write it every time, even if it already exists?

phmccarty: Yes. My intention was for autospec to own the file, but I can avoid overwriting the file if you think it should be user-controlled after it is first written.

matthewrsj: No I don't think it should be user controlled. We'll end up with different files checked in depending on who runs autospec. This is fine.

PR #165, opened 2018-05-30
Add detection of candidates for HSW technology

VictorRodriguez:
If a binary or .so is detecting to use basic MMX and SSE instructions
autospec will recomend to set use_avx2 = true in options.conf and
rebuild again

This will help developers to detect packages that are candidate to use
HSW technology such as AVX2 and FMA

Despite the fact that CLR has /usr/bin/avxjudge.py, others external ussers
might not be using autospec inside CLR.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -23,6 +23,7 @@ import shutil
 import re
 import tempfile
 import configparser
+import fileinput
 
 import build
 import buildpattern
@@ -320,6 +321,10 @@ def package(args, url, name, archives, workingdir, infile_dict):
                                                tarball.name,
                                                tarball.version,
                                                tarball.release)
+        if filemanager.is_avx_candidate(mock_chroot):
+            print("Autospec found files candidates to AVX technology !")
+            print("Is recomended to set use_avx2 = true in options.conf")
+
         if filemanager.clean_directories(mock_chroot):
             # directories added to the blacklist, need to re-run
             build.must_restart += 1
--- a/autospec/avxjudge.py
+++ b/autospec/avxjudge.py
@@ -1,0 +1,512 @@
+#!/usr/bin/python3
+import subprocess
+import sys
+import re
+import argparse
+import os
+
+# MMX and SSE2 instructions
+sse_instructions_xmm = list()
+
+# 0.1 value instructions
+avx2_instructions_lv = list()
+avx2_instructions_ymm = list()
+avx512_instructions_lv = list()
+
+
+# 1.0 value instructions
+avx2_instructions = list()
+avx512_instructions = list()
+
+# 2.0 value instructions
+avx2_instructions_hv = list()
+avx512_instructions_hv = list()
+
+
+
+
+sse_functions = dict()
+avx2_functions = dict()
+avx512_functions = dict()
+sse_functions_ratio = dict()
+avx2_functions_ratio = dict()
+avx512_functions_ratio = dict()
+
+
+verbose: int = 0
+quiet: int = 0
+
+
+
+
+def init_ins() -> None:
+
+
+    sse_instructions_xmm.append("paddb")
+    sse_instructions_xmm.append("paddd")
+    sse_instructions_xmm.append("paddsb")
+    sse_instructions_xmm.append("paddsw")
+    sse_instructions_xmm.append("paddusb")
+    sse_instructions_xmm.append("paddusw")
+    sse_instructions_xmm.append("paddw")
+    sse_instructions_xmm.append("pmaddwd")
+    sse_instructions_xmm.append("pmulhw")
+    sse_instructions_xmm.append("pmullw")
+    sse_instructions_xmm.append("psubb")
+    sse_instructions_xmm.append("psubsb")
+    sse_instructions_xmm.append("psubsw")
+    sse_instructions_xmm.append("psubusb")
+    sse_instructions_xmm.append("paddusw")
+    sse_instructions_xmm.append("paddw")
+    sse_instructions_xmm.append("pmaddwd")
+    sse_instructions_xmm.append("pmulhw")
+    sse_instructions_xmm.append("pmullw")
+    sse_instructions_xmm.append("psubb")
+    sse_instructions_xmm.append("psubd")
+    sse_instructions_xmm.append("psubd")
+    sse_instructions_xmm.append("psubsb")
+    sse_instructions_xmm.append("psubsw")
+    sse_instructions_xmm.append("psubusb")
+    sse_instructions_xmm.append("psubusw")
+    sse_instructions_xmm.append("psubw")
+
+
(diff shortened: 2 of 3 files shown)


autospec/autospec.py
```diff
@@ -320,6 +321,10 @@ def package(args, url, name, archives, workingdir, infile_dict):
                                                tarball.name,
                                                tarball.version,
                                                tarball.release)
+        if filemanager.is_avx_candidate(mock_chroot):
+            print("Autospec found files candidates to AVX technology !")
```
matthewrsj: This is poor grammar, maybe "Autospec found file candidates for AVX optimizations"

autospec/autospec.py
```diff
@@ -320,6 +321,10 @@ def package(args, url, name, archives, workingdir, infile_dict):
...
+            print("Is recomended to set use_avx2 = true in options.conf")
```
matthewrsj: Poor grammar again, maybe "Set use_avx2 to 'true' in options.conf to use features"

Also, "recomended" is spelled wrong.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
+#!/usr/bin/python3
+import subprocess
+import sys
+import re
+import argparse
+import os
+
+# MMX and SSE2 instructions
+sse_instructions_xmm = list()
```
matthewrsj: This entire file should be re-written into idiomatic python. I'll put suggestions in a few places below but I won't comment on everything.

First thing, this can definitely be refactored to not need so many globals.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+# 1.0 value instructions
+avx2_instructions = list()
+avx512_instructions = list()
+
+# 2.0 value instructions
+avx2_instructions_hv = list()
+avx512_instructions_hv = list()
+
+
+
```
matthewrsj: Why all the whitespace?

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+verbose: int = 0
+quiet: int = 0
+
+
+
+
+def init_ins() -> None:
+
+
+    sse_instructions_xmm.append("paddb")
```
matthewrsj: Do not append every single value... use `extend` instead.

Better than that, just set the `sse_instructions_xmm` list to a literal list like
```python
sse_instructions_xmm = [
    "one",
    "two",
    ...
]
```

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+import sys
+import re
+import argparse
+import os
+
+# MMX and SSE2 instructions
+sse_instructions_xmm = list()
+
+# 0.1 value instructions
+avx2_instructions_lv = list()
```
matthewrsj: Just for your own edification, not a big deal here, prefer using the literal object instead of the constructor. The call to `list()` or `dict()` here causes an extra lookup for that function while the interpreter immediately knows what to do with `[]` or `{}`... not a big deal though.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+
+sse_functions = dict()
+avx2_functions = dict()
+avx512_functions = dict()
+sse_functions_ratio = dict()
+avx2_functions_ratio = dict()
+avx512_functions_ratio = dict()
+
+
+verbose: int = 0
```
matthewrsj: Only going to comment on this once. We decided not to use these type identifiers in autospec. It restricts users to using 3.6 and we are at least supporting back to 3.5. Also it is inconsistent with the rest of the codebase.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+    avx2_instructions_hv.append("vfnmsub132ps")
+    avx2_instructions_hv.append("vfnmsub213ps")
+    avx2_instructions_hv.append("vfnmsub231ps")
+    avx2_instructions_hv.append("vfnmsub132pd")
+    avx2_instructions_hv.append("vfnmsub231pd")
+    avx2_instructions_hv.append("vfnmsub213pd")
+    avx2_instructions_hv.append("vdivpd")
+
+
+    return
```
matthewrsj: return not needed.

Also, if all this function is doing is setting the values for these instruction lists, this is a case where you could use a global and just set it to the literal list of values.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+        val = max(val, 0.1)
+    if instruction in avx2_instructions:
+        val = max(val, 1.0)
+    if instruction in avx2_instructions_hv:
+        val = max(val, 2.0)
+
+    return val
+
+def has_high_register(args: str) -> int:
+    if "mm16" in args:
```
matthewrsj: Ok, this is not how python works. Here's a little lesson
```python
ls = [1, 2, 3, 4]
if any(l in ints for l in ls):
    return True
```
Also, use a bool! They exist natively in Python!

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+    if "mm30" in args:
+        return 1
+    if "mm31" in args:
+        return 1
+    return 0
+
+def is_avx512(instruction:str, args:str) -> float:
+    val: float = -1.0
+
+    if instruction in avx512_instructions_lv:
```
phmccarty: `avx512_instruction_lv` is initialized to an empty list but is not populated anywhere, so this condition looks like it will always be False...

Is this an oversight, or is it reserved for future expansion of the code?

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+        val = max(val, 0.1)
+    if instruction in avx512_instructions:
```
phmccarty: Same comment as above, but for `avx512_instructions`.

autospec/avxjudge.py
```diff
@@ -0,0 +1,512 @@
...
+        val = max(val, 1.0)
+    if instruction in avx512_instructions_hv:
```
phmccarty: Same comment as above, but for `avx512_instructions_hv`.

autospec/avxjudge.py
phmccarty: In the context of this code, I agree that it makes more sense for these lists to be initialized in global scope (at the top of the source file), and no `init_ins()` call would then be needed.

VictorRodriguez:
@phmccarty @matthewrsj thanks a lot for your feedback I definitely will not ship the avxjudge on the autospec project, i will rather prefer to link to the one provided by clear in : https://github.com/clearlinux-pkgs/clr-python-timestamp/blob/master/clr-python-timestamp.spec#L61
I still don't know how, but is better than maintain 2 places. I will cancel the PR adn came back with something way much better 

regards

PR #167, opened 2018-06-11
Set default permissions for doc files

icarus-sparry:
Make the generated spec file set the default attributes for doc files
to be 0644, so the documentation is not executable.

Signed-off-by: Icarus Sparry <icarus.w.sparry@intel.com>

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -328,7 +328,10 @@ class Specfile(object):
                 continue
 
             self._write("\n%files {}\n".format(pkg))
-            self._write("%defattr(-,root,root,-)\n")
+            if pkg in ["doc"]:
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -328,7 +328,10 @@ def write_files(self):
                 continue
 
             self._write("\n%files {}\n".format(pkg))
-            self._write("%defattr(-,root,root,-)\n")
+            if pkg in ["doc"]:
```
bryteise: Why not pkg == "doc"?

icarus-sparry: In case we want to add the debug info shortly

bryteise: I'll look forward to it.

bryteise:
Looks good to me.

PR #169, opened 2018-06-13
Port whatrequires generation to DNF

### LINKED ISSUE #164: Port whatrequires support to DNF
The `whatrequires` file creation currently use YUM and repoquery from yum-utils. Because most Clear Linux tooling has moved to DNF now, autospec should do the same.

phmccarty:
Fixes #164

Because most of Clear Linux OS tooling has switched over to DNF, autospec should also use DNF for whatrequires generation.


bryteise:
Yaay! Looks good.

PR #171, opened 2018-06-14
Various autospec improvements

fenrus75:


bryteise:
Works for me. I await the template solution from @icarus-sparry so we don't keep needing this huge clump of things in specfile.py though.

PR #172, opened 2018-06-14
Fix handling of symlinks to directories in %files

### LINKED ISSUE #135: Confusing warning printed regarding directory removal
When I rerun autospec on the Clear Linux `groff` package (https://github.com/clearlinux-pkgs/groff), autospec prints a confusing warning message where `[WARNING]` is highlighted in red:

```
Building package groff round 1
groff mock chroot at /var/lib/mock/clear-groff
  Adding buildreq: bison
  Adding buildreq: texinfo
  New %files content found
[WARNING] Removing directory /usr/share/groff/current from file list
Building package groff round 2
groff mock chroot at /var/lib/mock/clear-groff
RPM build successful
...
```

It's unclear to me how to respond to this warning. The directory clearly disappears from the %files section, but why is it removed in the first place?

phmccarty:
Fixes #135 

Because os.path.isdir() follows symlinks, if it is passed a symlink to a directory, the symlink will be removed from the %files section here.

Ensure the removal does not occur for this case by making the conditional check more strict.

### DIFF
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -145,7 +145,8 @@ class FileManager(object):
                 res.add(f)
                 continue
 
-            if os.path.isdir(os.path.join(root, f.lstrip("/"))):
+            path = os.path.join(root, f.lstrip("/"))
+            if os.path.isdir(path) and not os.path.islink(path):
                 util.print_warning("Removing directory {} from file list".format(f))
                 self.files_blacklist.add(f)
                 removed = True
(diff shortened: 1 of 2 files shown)


bryteise:
Oh nice change. Do we have any tests that we could add to for ensuring we don't break this in the future?

phmccarty:
Yes, I will add a new unit test for this.

PR #176, opened 2018-06-21
Move srpms and debuginfo into their own folder

tmarcu:
Debug info rpms are not added to bundles, and adding source RPMs will
make mixer crash because it looks for files that will not exist. Break
up the RPMs so it is more obvious which RPMs should be copied over for
building an image, and which ones are useful meta-data.

Signed-off-by: Tudor Marcu <tudor.marcu@intel.com>

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -240,3 +240,18 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
         exit(1)
 
     parse_build_results(download_path + "/results/build.log", returncode, filemanager)
+
+    files = os.listdir('{}/results'.format(download_path))
+
+    os.makedirs('{}/results/srpm'.format(download_path), exist_ok=True)
+    os.makedirs('{}/results/debuginfo'.format(download_path), exist_ok=True)
+    os.makedirs('{}/results/logs'.format(download_path), exist_ok=True)
+
+    # organize output files so it's more clear what to use for mixes
+    for f in files:
+        if f.endswith(".src.rpm"):
+            shutil.move('{}/results/{}'.format(download_path, f), '{}/results/srpm/{}'.format(download_path, f))
+        if "debuginfo" in f:
+            shutil.move('{}/results/{}'.format(download_path, f), '{}/results/debuginfo/{}'.format(download_path, f))
+        if f.endswith(".log"):
+            shutil.move('{}/results/{}'.format(download_path, f), '{}/results/logs/{}'.format(download_path, f))
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -126,7 +126,7 @@ def save_build_log(path, iteration):
     Must be saved outside of the results/ directory since it gets wiped away on
     each round.
     """
-    buildlog = os.path.join(path, "results", "build.log")
+    buildlog = os.path.join(path, "results", "logs", "build.log")
     shutil.copyfile(buildlog, "{}/build.log.round{}".format(path, iteration))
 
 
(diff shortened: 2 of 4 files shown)


tmarcu:
$ tree results/
results/
├── build.log
├── debuginfo
│   └── mixer-tools-debuginfo-4.6.5-92.x86_64.rpm
├── mixer-tools-4.6.5-92.x86_64.rpm
├── mixer-tools-bin-4.6.5-92.x86_64.rpm
├── mixer-tools-data-4.6.5-92.x86_64.rpm
├── mixer-tools-license-4.6.5-92.x86_64.rpm
├── mixer-tools-man-4.6.5-92.x86_64.rpm
├── root.log
├── srpm
│   └── mixer-tools-4.6.5-92.src.rpm
└── state.log

autospec/build.py
```diff
@@ -240,3 +240,16 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
...
+
+    files = os.listdir('{}/results'.format(download_path))
+
+    os.makedirs('{}/results/srpm'.format(download_path))
+    os.makedirs('{}/results/debuginfo'.format(download_path))
+
+    # organize output files so it's more clear what to use for mixes
+    for f in files:
+        if ".src.rpm" in f:
+            print('copying {} to srpm')
```
matthewrsj: Autospec's output is already extremely spammy, mind taking this out?

autospec/build.py
```diff
@@ -240,3 +240,16 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
         exit(1)
 
     parse_build_results(download_path + "/results/build.log", returncode, filemanager)
+
+    files = os.listdir('{}/results'.format(download_path))
+
+    os.makedirs('{}/results/srpm'.format(download_path))
```
matthewrsj: What about when someone already has these directories on their system from a previous run? This will fail with a traceback. You should capture `FileExistsError` exceptions and ignore them.

autospec/build.py
tmarcu: Oh yeah totally ^

autospec/build.py
tmarcu: I wrongly assumed it makedirs works like mkdir -p :) fixing.

tmarcu:
@bryteise @phmccarty Please chime in on any corresponding tooling changes that may be affected by this, thanks!

bryteise:
I only know of https://github.com/clearlinux/common/blob/master/Makefile.common#L15

matthewrsj:
DO NOT MERGE until https://github.com/clearlinux/common/pull/5/files is merged!

autospec/build.py
```diff
@@ -240,3 +240,18 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
...
+    os.makedirs('{}/results/debuginfo'.format(download_path), exist_ok=True)
+    os.makedirs('{}/results/logs'.format(download_path), exist_ok=True)
+
+    # organize output files so it's more clear what to use for mixes
+    for f in files:
+        if ".src.rpm" in f:
+            shutil.move('{}/results/{}'.format(download_path, f), '{}/results/srpm/{}'.format(download_path, f))
+        if "debuginfo" in f:
+            shutil.move('{}/results/{}'.format(download_path, f), '{}/results/debuginfo/{}'.format(download_path, f))
+        if ".log" in f:
```
phmccarty: There is a package in Clear Linux named `R-futile.logger`, so with this change, RPMs for that package will be copied into `logs` by mistake. Better to check for a `.log` file suffix here instead.

PR #179, opened 2018-06-26
Add buildreq_cache

bryteise:
Enable having a cache for detected buildreqs to avoid needing to
redetect the same buildreqs when rebuilding the same version of a
package.


phmccarty:
Testing this code, I see that if autospec detects no new additional build requirements, the `buildreq_cache` file is created containing only the version number, and it's not committed. Should `buildreq_cache` even be created in this case?

bryteise:
Updated to avoid creating the file when no buildreqs are added to the cache and to add the file to config_files so that they are commited to git when they are created without first existing.

PR #181, opened 2018-07-02
Use the new way of doing AVX2 and AVX512 packages

thiagomacieira:
The %make_install_avx2 and %make_install_avx512 macros were added to
clr-rpm-config release 85 and will take care of moving the files to the
correct place, without the need to pass --libdir or --bindir to
configure. Passing those options could result in some packages being
broken, as they could search for plugins in /usr/lib64/haswell/foopkg,
which usually won't exist.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -544,12 +544,12 @@ class Specfile(object):
 
         if config.config_opts['use_avx2']:
             self._write_strip("pushd ../buildavx2/")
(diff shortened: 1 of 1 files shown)


bryteise:
Similar to the other commit, I'll wait a bit to merge this to let folks get a new enough clear version.

bryteise:
Really happy with the direction of this cleanup to the code base btw =).

PR #182, opened 2018-07-02
cmake: properly unset the PKG_CONFIG_PATH variable

thiagomacieira:
Otherwise, it'll be used by the AVX2 and AVX512 builds. The CFLAGS and
CXXFLAGS variables will be reset by the write_variables() calls.


thiagomacieira:
No preference, but this way seemed cleaner: unset where you set it.

PR #184, opened 2018-07-04
Start printing failed patterns

fenrus75:
Missing failed patterns has led to packages having subtle bugs; start printing all missing patterns


bryteise:
Updated via #186

PR #192, opened 2018-07-17
Add Module::Build as buildreq for Perl packages

### LINKED ISSUE #191: Fails to build Perl packages which use Module::Build::Compat
Some Perl packages use Module::Build::Compat to provide both Build.PL and Makefile.PL. autospec attempts Makefile.PL first, then falls back to Build.PL. For such packages, Makefile.PL requires Module::Build, but the buildreq is not added automatically because the dependency on Module::Build is not detected from the Makefile.PL output.

I have two potential fixes to offer, either of which should be sufficient:
- Add a pattern to simple_pats to match the failure string from Makefile.PL and add perl(Module::Build) to buildreqs
- Switch the order in write_cpan_pattern to attempt Build.PL before Makefile.PL (because the dependency is correctly detected when executing Build.PL)

Packages that only provide Build.PL seem to work fine, as do packages that provide only Makefile.PL and don't require Module::Build.

bwarden:
Add perl(Module::Build) to buildreqs for Perl packages that use
Module::Build to build themselves.

Fixes #191

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -598,6 +598,7 @@ def scan_for_configure(dirn):
             buildpattern.set_build_pattern(python_pattern, default_score)
 
         if "Makefile.PL" in files or "Build.PL" in files:
+            add_buildreq("perl-Module-Build")
             buildpattern.set_build_pattern("cpan", default_score)
 
         if "SConstruct" in files:


bryteise:
Thanks!

fenrus75:
-1
this is the wrong thing to add
the right thing to add is buildreq-cpan
which pulls all perl reqs in

bryteise:
Ahh right, I forgot we just added those. I'll send an update.

PR #193, opened 2018-07-17
switch to using buildreq-* meta packages for build deps

fenrus75:


bryteise:
@fenrus75 I updated the patch and tested a few python things in #194. Also fixed the build error if you want to look that over.

Issue #189, opened 2018-07-12
Misparsing License: field from upstream tarball's included spec file


bwarden:
Upstream tarball https://github.com/mozilla/rr/archive/5.2.0.tar.gz contains a spec file, in which:
```
License:        @CPACK_RPM_PACKAGE_LICENSE@
```

Autospec attempts to interpret this as a valid license specifier, resulting in adding "@CPACK_RPM_PACKAGE_@" (note "LICENSE" trimmed) to the License field in the spec file it generates.

bryteise:
Fixed via #197

PR #197, opened 2018-07-18
Be more picky on licenses from included spec files

bryteise:
Exclude likely macros from being used as licenses autospec will pick
up from the spec files included in the project's sources. As part of
this stop potentially adding urls as license strings.


bryteise:
Fixes #189

PR #200, opened 2018-07-19
Extract ARCHIVES with setup macro

bryteise:
Move to extracting archives with the setup macro instead of hardcoding
a tar extraction which fails for zip files. As part of this change,
adapt how content is being moved from the extracted archive to the
destination folder.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -381,10 +381,6 @@ class Specfile(object):
     def write_prep(self, ruby_pattern=False):
         """Write prep section to spec file"""
         self._write_strip("%prep")
-        for archive in self.sources["archive"]:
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -394,15 +390,18 @@ def write_prep(self, ruby_pattern=False):
                 self._write_strip("%setup -q -c -n " + self.tarball_prefix)
             else:
                 self._write_strip("%setup -q -n " + self.tarball_prefix)
+                for archive in self.sources["archive"]:
+                    self._write_strip("%setup -D -T -n {0} -a {1}"
```
phmccarty: Please pass the `-q` option too, so that the tarball/zip extraction does not spam the build log.

bryteise:
@phmccarty Updated switching to -b instead of -a and added -q.

PR #202, opened 2018-07-20
Enforce zipfile prefix to be valid

bryteise:
Autospec doesn't allow tarfiles to be missing a prefix so the same
should be done for zipfiles.


bryteise:
Yep, just a hardcoded string to fight with =P.

Issue #203, opened 2018-07-20
Version number for graphviz package incorrectly generated.


kevincwells:
Starting 6 weeks ago (06-08-2018), the version number for the `graphviz` package is being extracted/detected incorrectly. The correct version is `2.40.1`, but it is being reported as the default `1`.

bryteise:
Fixed in #204

phmccarty:
With the change from @bryteise, the graphviz detected version is now `stable.2.40.1`, which is better. We can iterate on this.

kevincwells:
I'm sure this isn't the best place to document this, but for lack of an easier option, the following (non-clear-specific) packages have version numbers that don't match upstream:



|Package|Lost track of Version| Comments|
|-------|---------------------|---------|
| beignet | 19 mo ago | |
| lshw | always | |
| mesa | 4 mo ago | Already past the version # of last known CVE by time lost track |
| pcre | 16 mo ago | Lost track while on v8.41 which had unpatched CVE. Manual inspection shows we upgraded to v8.42 four months ago, but unclear how to detect this in automated fashion |
| unzip | 4 mo ago | Lost track while on v6.0, with several unpatched CVEs. Appear to still be on v6.0. |

PR #204, opened 2018-07-20
Add gitlab parsing for name and version

bryteise:
Add specific parsing for gitlab urls in order to derive the name and
version information for a package.


bryteise:
Fixes #203

PR #207, opened 2018-07-25
Let make_check_command work for meson

cmarcelo:
In the future would be nice to auto-identify this in scan_for_tests,
but the opt-in (by adding make_check_command file) is already useful.


thiagomacieira:
Looks good, but I haven't tested.

PR #209, opened 2018-07-25
count: support counting test results from meson

cmarcelo:
This patch takes a different approach, since it was likely that
changing the way we count the existing individual patterns could break
other logs.

So when it identifies this is a meson test, it forks off to just parse
the meson patterns.

It doesn't support the "CLR-XTEST:" prefix to group multiple tests,
mostly because I couldn't find examples of it to test.

### DIFF
--- a/autospec/count.py
+++ b/autospec/count.py
@@ -129,6 +129,25 @@ def convert_int(intstr):
         return 0
 
 
+def parse_meson_test(lines):
+    global total_pass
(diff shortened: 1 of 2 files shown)


cmarcelo:
Relates to #207. Without this, the counting was giving wrong results.

I can update the patch to play nice with "CLR-XTEST" if necessary (just let me know where we use it so we can test).

PR #210, opened 2018-07-28
Add a parser for CMake instruction files

thiagomacieira:
This is the first part of the change. It will now scan all
CMakeLists.txt and *.cmake files for find_package() calls and match
those against the cmake_modules file.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -308,6 +308,45 @@ def rakefile(filename):
                 print("Rakefile-new: rubygem-" + s)
 
 
+def parse_cmake(filename):
+    """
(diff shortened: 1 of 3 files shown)


autospec/buildreq.py
```diff
@@ -308,6 +308,35 @@ def rakefile(filename):
...
+                add_buildreq(pkg)
+            except:
+                pass
+
+        match = pkgconfig.search(line)
+        if match:
+            for s in re.split(r'\s+', match.group(1)):
+                if s in pkg_search_modifiers:
+                    continue
+                add_pkgconfig_buildreq(s)
```
thiagomacieira: This is adding deps like `pkgconfig("systemd")` (with the quotes). Investigating.

PR #212, opened 2018-07-31
Improve autospec patterns

bryteise:


bryteise:
Ah good point, updated.

Issue #213, opened 2018-07-31
Mis-parsing strange requirements specification


bwarden:
autospec misunderstood this content from setup.py in flake8-import-order:
```
    install_requires=[
        "enum34 ;  python_version <= '2.7'",
        "pycodestyle",
        "setuptools",
    ],
```

```
Processing https://files.pythonhosted.org/packages/5b/5b/fd248ea91880a7b5e4754f396f4598e8244f28df0d0f8790453acbafc7c4/flake8-import-order-0.18.tar.gz
=========================================================================================================
Name        : flake8-import-order
Version     : 0.18
Prefix      : flake8-import-order-0.18
Banning build requirement: enum34 ;  python_version.
requirement 'enum34 ;  python_version' not found is buildreqs or os_packages, skipping
requirement '[:python_version' not found is buildreqs or os_packages, skipping
Buildreqs   : buildreq-distutils3 pycodestyle setuptools
Adding license from PKG-INFO: LGPLv3
Adding license from PKG-INFO: LGPLv3
Summary     : Flake8 and pylama plugin that checks the ordering of import statements.
License     :  LGPL-3.0  (server) ( f45ee1c765646813b442ca58de72e20a64a7ddba )
Licenses    :  LGPL-3.0
```

...and it still puts it in the specfile:
```
Requires: enum34 ;  python_version
BuildRequires : enum34 ;  python_version
```

PR #214, opened 2018-07-31
Improve python buildreq detection

bryteise:
Add additional logic to handle: "pkg ;  dep <= 'X'", style statements
for python requirements in setup.py. Also update tests to validate the
results are as expected.


bryteise:
Fixes #213

Issue #177, opened 2018-06-26
Archive all results for each build round


phmccarty:
At the moment, autospec archives just the mock build.log for each build round. I would like to see autospec expand to archiving *all* results from each build round (i.e. root.log, build.log, state.log, mock_*.log, and the srpm).

Implementation idea: Pass `--result=results-buildN` to mock for each build round `N`, and rename the final `results-buildN` directory to `results`. Or maybe symlink `results` to `results-buildN` for the final `N`.

PR #217, opened 2018-08-01
Updates for Mock logs handling and common tooling integration

phmccarty:
The Clear Linux common tooling has some pending changes to how Mock logs are handled, reverting mostly to the previous behavior, but introducing some improvements as well.

This PR aligns autospec with those changes.


bryteise:
Fixes #177

PR #226, opened 2018-08-16
Add %check rules for AVX2, AVX512 and 32-bit builds too

thiagomacieira:


thiagomacieira:
ping

PR #229, opened 2018-08-17
Fix parsing of CMake files: only do it if the we're using CMake

thiagomacieira:


thiagomacieira:
Yeah, took me a while to notice. The difference is the score: if the CMakeLists.txt file is not in the root directory, the score is 1, so the build pattern may be different than cmake.

bryteise:
Ah nice, no grabbing cmake bits when not using cmake.

PR #230, opened 2018-08-17
Add the ability to move certain files to dev

thiagomacieira:
This is useful when the package includes executables used for
development but not for other cases.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -696,12 +696,17 @@ def parse_config_files(path, bump, filemanager, version):
 
     content = read_conf_file(os.path.join(path, "extras"))
     for extra in content:
-        print("extras for: %s." % extra)
+        print("extras for  : %s." % extra)
     filemanager.extras += content
+
+    content = read_conf_file(os.path.join(path, "dev_extras"))
+    for extra in content:
+        print("dev for     : %s." % extra)
+    filemanager.dev_extras += content
 
     content = read_conf_file(os.path.join(path, "setuid"))
     for suid in content:
-        print("setuid for: %s." % suid)
+        print("setuid for  : %s." % suid)
     filemanager.setuid += content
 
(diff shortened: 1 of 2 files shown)


fenrus75:
more of a policy question
is the case of moving all /usr/bin to -dev or is it broader than that.

we have a flag already for moving all .so out of dev ; if the case really
is just bin, we should consider (also) doing a boolean flag.
boolean flags in options.conf are a bit more user friendly in some ways
(but more restrictive obviously)

if it's not the simple case then yeah we need the full more complex case
like this patch does

(they're higher maintenance than just the boolean in the sense that these
files don't support wildcards so more fragile)

thiagomacieira:
For qtbase, it is all of bin, but that is not necessarily the case for all packages. For example, for qttools, the following remain in the main (-bin) package:

 /usr/bin/assistant
 /usr/bin/qdbus
/usr/bin/qdbusviewer
 /usr/bin/qtdiag
 /usr/bin/qtpaths

bryteise:
Given the requirements of it being a subset of /usr/bin I'll take the patch as is.

thiagomacieira:
Sorry, qtbase build just finished and this wasn't sufficient. I managed to move things from the main package to -dev, but extras doesn't move things out of the lib package. There's more work needed.

autospec/config.py
```diff
@@ -696,12 +696,17 @@ def write_default_conf_file(name, description):
 
     content = read_conf_file(os.path.join(path, "extras"))
     for extra in content:
-        print("extras for: %s." % extra)
+        print("extras for  : %s." % extra)
     filemanager.extras += content
 
+    content = read_conf_file(os.path.join(path, "dev_extras"))
```
cmarcelo: Please document the new file ("dev_extras") in README.rst.

thiagomacieira:
See #235

PR #231, opened 2018-08-17
Fix the location of files explicitly moved by configuration

thiagomacieira:
Instead of adding an %exclude line to the main package, simply don't
list it at all. RPM only complains if the file isn't in any package at
all; since we're adding to one of the subpackages, that is sufficient.

This fixes the issue that it wasn't possible to explicitly move a file
moved from one subpackage to another (dev or extras), as the
pattern-matching would make it show up in two.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -711,11 +711,8 @@ def parse_config_files(path, bump, filemanager, version):
 
     content = read_conf_file(os.path.join(path, "attrs"))
     for line in content:
-        attr = re.split(r'\(|\)|,', line)
-        attr = [a.strip() for a in attr]
-        filename = attr.pop()
         print("attr for: %s." % filename)
-        filemanager.attrs[filename] = attr
+    filemanager.attrs += content
 
     patches += read_conf_file(os.path.join(path, "series"))
     pfiles = [("%s/%s" % (path, x.split(" ")[0])) for x in patches]
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -25,6 +25,7 @@ import config
 import re
 import os
 import util
+import sys
 from collections import OrderedDict
 # todo package splits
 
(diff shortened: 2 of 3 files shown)


thiagomacieira:
I see what you mean. I can leave the %excludes, but keep these new `return`. I'll clean up the wildcards if I can in another commit.

autospec/files.py
```diff
@@ -86,17 +221,15 @@ def compat_exclude(self, filename):
 
         return exclude
 
-    def file_pat_match(self, filename, pattern, package, replacement="", prefix=""):
+    def file_pat_match(self, filename, pat, package, prefix=""):
```
bryteise: Nit around keeping pattern as the variable name (though it is a pain with patterns used as a name in the same block).

autospec/config.py
```diff
@@ -711,11 +711,8 @@ def write_default_conf_file(name, description):
 
     content = read_conf_file(os.path.join(path, "attrs"))
     for line in content:
-        attr = re.split(r'\(|\)|,', line)
-        attr = [a.strip() for a in attr]
-        filename = attr.pop()
         print("attr for: %s." % filename)
-        filemanager.attrs[filename] = attr
+    filemanager.attrs += content
```
bryteise: So I'm inclined to just delete this entirely as it isn't used, it is pretty old code and more headache than value when looking at this change.

thiagomacieira: I'll split the removal off to a new PR and rebase.

autospec/files.py
thiagomacieira: I renamed here because the local variable of the compiled object was `pat`. Do you want me to change the internal code but keep the variable name in the function declaration?

bryteise: Only because I prefer full names in function declarations though I don't think we follow that super well in autospec today.

autospec/config.py
bryteise: Thanks!

thiagomacieira:
Oops, didn't mean to close

autospec/files.py
```diff
@@ -41,14 +42,148 @@ def __init__(self):
...
+            if file.startswith("/") or file.startswith('%'):
+                return (re.compile(re.escape(file)), package)
+            if file.startswith("^"):
+                return (re.compile(file), package)
+            util.print_fatal("Line in {} pattern must be absolute or start with ^: {}"
+                             .format(package, file))
+            sys.exit(1)
+
+        # Prepare compiled patterns
+        self.patterns = []
```
bryteise: self.patterns should be initialized in __init__

autospec/files.py
```diff
@@ -184,144 +317,24 @@ def push_file(self, filename):
...
-            self.excludes.append(filename)
-        if filename in self.dev_extras:
-            self.push_package_file(filename, "dev")
-            self.excludes.append(filename)
-
-        if filename in self.setuid:
-            newfn = "%attr(4755, root, root) " + filename
-            self.push_package_file(newfn, "setuid")
-            self.excludes.append(filename)
+            self.push_package_file("%exclude " + filename)
```
bryteise: I almost feel like we'd like an excludes flag for a push_package_file so the syntax for this is consolidated. I need to think about it more.

autospec/files.py
```diff
@@ -184,144 +317,24 @@ def push_file(self, filename):
...
-            (r"^/usr/lib/sysctl.d", "config"),
-            (r"^/usr/share/", "data"),
-            # finally move any dynamically loadable plugins (not
-            # perl/python/ruby/etc.. extensions) into lib package
-            (r"^/usr/lib/.*/[a-zA-Z0-9\.\_\-\+]*\.so", "lib"),
-            (r"^/usr/lib64/.*/[a-zA-Z0-9\.\_\-\+]*\.so", "lib"),
-            (r"^/usr/lib32/.*/[a-zA-Z0-9\.\_\-\+]*\.so", "lib32"),
-            # locale data gets picked up via file_is_locale
-            (r"^/usr/share/locale/", "ignore")]
+        if self.want_dev_split and self.file_pat_match(filename, re.compile(r"^/usr/.*/include/.*\.h$"), "dev"):
```
bryteise: Since we have a nice little object we could initialize these static regex strings there, not required for this commit though but maybe a todo for later.

bryteise:
Alrighty first pass review completed, sorry that was in parts X(. The idea is definitely what we want and is pretty much there.

phmccarty:
If I understand correctly, the changes implemented with the excludes rework (#435) makes this PR unnecessary. Files that were formerly `%exclude`d via the `excludes` control file are now removed with `rm -f ...` in the `%install` section. Closing.

PR #238, opened 2018-08-27
Add 7z support

bwarden:
The mkdir -p change is required for archives that need to be pre-extracted under the build directory for cmake projects.

The main change is implementing a function to handle 7z archives, basically ripped off from build_unzip, but modified to deal with 7z's extra-verbose headers.

All together, this allows me to build a cmake-based package that requires downloading some media and translations that are distributed as 7z archives.


bryteise:
Would you mind adding some testing for the build_un7z function? You can look at tests/test_tarball.py for the build_unzip that already exists for the kind of test to write.

PR #240, opened 2018-08-30
Document handling of systemd configuration template files

bryteise:


README.rst
```diff
@@ -503,3 +503,12 @@ URL. The implementation should show the now-stored license file via a
 web page, and enable a human to make a decision on the license. This is
 then stored internally, allowing future requests to automatically know
 the license type when this hash is encountered again.
+
+
+Integration of systemd unit files
+=================================
+``autospec`` can add most systemd template file types by having a file in the
+filename.extension in the build directory. Supported extensions are:
```
phmccarty: Should we describe the filename as `$pkgname.extension` instead, for consistency with descriptions elsewhere?

phmccarty: Err, I meant `$package.extension`

bryteise: That's a good point but $package.extension isn't quite right here because these files are not always going to follow that naming convention and I don't want to convey that files without the package name but the right extension won't be used. Thoughts?

PR #245, opened 2018-09-05
Fix cmake pkgconfig whitespace

bwarden:
cmake content like `PKG_CHECK_MODULES(PC_FFTW3F "fftw3f >= 3.0")` resulted in a failed attempt to add `pkgconfig(fftw3f )`. Add a unit test and logic to handle this.


tests/test_buildreq.py
```diff
@@ -450,6 +450,20 @@ def test_parse_cmake_pkg_check_modules(self):
         self.assertEqual(buildreq.buildreqs,
                          set(['pkgconfig(gio-unix-2.0)', 'pkgconfig(glib-2.0)']))
 
+    def test_parse_cmake_pkg_check_modules_whitespace(self):
+        """
+        Test parse_cmake to ensure accurate detection of versioned and
+        unversioned pkgconfig modules.
```
phmccarty: Though it's pretty clear what this test's purpose is, can you revise the docstring to more explicitly describe what it covers? It appears to be a copy/paste of the `test_parse_cmake_pkg_check_modules` docs at the moment.

PR #246, opened 2018-09-06
Improve handling variables in pkg_check_modules

bwarden:
Strip out variables in e.g.:
pkg_check_modules(AVCODEC libavcodec${_avcodec_ver})

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -181,7 +181,7 @@ def parse_modules_list(modules_string, is_cmake=False):
     requirements
     """
     if is_cmake:
-        modules = [m for m in re.split(r'(\s*[><]?=\s*)', modules_string)]
+        modules = [m for m in re.split(r'\s*([><]?=|\${?[^}]*}?)\s*', modules_string)]
         modules = filter(None, modules)
     else:
         modules = [m.strip('[]') for m in modules_string.split()]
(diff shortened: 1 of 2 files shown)


autospec/buildreq.py
```diff
@@ -181,7 +181,7 @@ def parse_modules_list(modules_string, is_cmake=False):
     requirements
     """
     if is_cmake:
-        modules = [m for m in re.split(r'(\s*[><]?=\s*)', modules_string)]
+        modules = [m for m in re.split(r'\s*([><]?=|\$\{[^}]*\})\s*', modules_string)]
```
bryteise: I've seen packages where the brackets are not part of the variable string, could you make them optional here? As a side note I'm starting to become worried about how long this regex will be something I understand X(.

bwarden: Let's see what I can do. It certainly won't make the regex shorter, though.

bryteise: I'll this change as is if it doesn't look like something you'd want to add, easy enough to buildreq_ban in the few places it comes up.

bwarden: Taking out two extraneous \ left me room to add two ?, so at least it's the same length :)

bryteise: Heh that's good enough for me.

bryteise: Mind updating the test to add a no bracket version too?

bwarden: I already did :)

bryteise: Hah I even looked before I asked and I just missed the $ in the second string, darn it X(. Thanks so much!

PR #248, opened 2018-09-08
More finer grained build control

thiagomacieira:
Improves `build_prepend` and adds `make_prepend`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -250,16 +250,21 @@ make32_install_args
 
 prep_prepend
   Additional actions that should take place directly after ``%prep``
-  This will be placed in the resulting ``.spec``, and is used for
-  situations where fine-grained control is required.
+  and before the ``%setup`` macro.  This will be placed in the
+  resulting ``.spec``, and is used for situations where fine-grained
+  control is required.
 
 build_prepend
   Additional actions that should take place directly after ``%build``
-  This will be placed in the resulting ``.spec``, and is used for
-  situations where fine-grained control is required.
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -250,16 +250,21 @@ make32_install_args
...
 
 build_prepend
   Additional actions that should take place directly after ``%build``
-  This will be placed in the resulting ``.spec``, and is used for
-  situations where fine-grained control is required.
+  and before the ``%configure`` macro or equivalent (``%cmake``,
+  etc.). If autospec is creating AVX2, AVX-512 or 32-bit, these
+  actions will be repeated for each of those builds, This will be
+  placed in the resulting ``.spec``, and is used for situations where
+  fine-grained control is required.
```
bryteise: Should we add a different version of build_prepend that gets added to each build type instead since this is modifying already existing behavior?

README.rst
```diff
@@ -262,6 +263,14 @@ build_prepend
...
   fine-grained control is required.
 
+make_prepend
+  Additional actions that should take place directly after the
+  configuring step and before the ``%make`` macro or equivalent. If
+  autospec is creating AVX2, AVX-512 or 32-bit, these actions will be
+  repeated for each of those builds, before their respective make
+  steps. This will be placed in the resulting ``.spec``, and is used
+  for situations where fine-grained control is required.
+
```
bryteise: Similar comment here as to if we should have a single action variant of this.

thiagomacieira:
I don't think there's a point of a single-action variant of this one. It would be almost always wrong.

PR #250, opened 2018-09-11
Make sub-packages require/provide exact versions

thiagomacieira:
I'm seeing at least one package that weirdly causes an old version to
get installed by dnf, causing file conflicts later.


bryteise:
This would be handy for testing purposes.

thiagomacieira:
What would be handy? Installing different versions? Or forcing the version?

bryteise:
Your change here for forcing the version. We sometimes run into problems with testing where dependencies are satisfied by an earlier version of a subpackage that isn't provided in the new release. This would break the link on that dependency in some cases making a more obvious failure rather than sneaking in a different version of a subpackage that won't be available when making a release.

bryteise:
I'd take it, thanks!

PR #252, opened 2018-09-15
Add detection of candidates for HSW technology

VictorRodriguez:
If a binary or .so is detecting to use basic MMX and SSE instructions
autospec will recomend to set use_avx2 = true in options.conf and
rebuild again

This will help developers to detect packages that are candidate to use
HSW technology such as AVX2 and FMA

This PR needs : https://github.com/clearlinux/clr-avx-tools/pull/8
to run 

### DIFF
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -25,6 +25,9 @@ import config
 import re
 import os
 import util
+import sys
+sys.path.append('/usr/share/clr-avx-tools/')
+from avxjudge import do_file
 from collections import OrderedDict
 # todo package splits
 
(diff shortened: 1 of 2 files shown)


autospec/files.py
```diff
@@ -25,6 +25,9 @@
 import re
 import os
 import util
+import sys
+sys.path.append('/usr/share/clr-avx-tools/')
```
thiagomacieira: Can we do this with a binary installed in /usr/bin instead?

autospec/files.py
```diff
@@ -25,6 +25,9 @@
...
+from avxjudge import do_file
```
thiagomacieira: I don't like this, you're importing another project's Python code. What if I rewrote avxjudge in Perl?

Can you modify avxjudge instead to output the information you want in a parseable format, which you can read and interpret here? And preferably use one of the wrappers in /usr/bin instead of relying on the internal script.

fenrus75:
also while I appreciate that we want to suggest this... I'm not a fan.
we need to do this ONLY for performance sensitive pieces where we can show
gains, not for the whole distro.

we can also just run the tool at the end of "make autospec" without having
to patch autospec for this

VictorRodriguez:
thanks for the feedback, I think that approach suggested by @fenrus75 to run this tool at the end of make autospec is the best approach. Now, regarding to the point of we need to do this ONLY for the performance sensitive places, I am not sure how to detect this w/o running this tool first, maybe we could add a flag file and if detected makeautospec run the avxjudge and generate the report. What do you recommend @fenrus75 and @thiagomacieira  ?

autospec/files.py
VictorRodriguez: sorry, this was a POC. I will use the avxjudge as a binary instead at the end of make autospec

phmccarty:
I think the consensus was that we did not want to couple these avxjudge checks with autospec. If this functionality is still desired, please continue the discussion elsewhere.

VictorRodriguez:
+ 1

Issue #254, opened 2018-09-17
options.conf allow_test_failures = true not always respected


jurobystricky:
autospec fails (for example, when running in perl-Feed-Find) due to some failing tests with an error:

RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.X9gxFI (%check)

This in spite of options.conf containing:
"allow_test_failures = true"

The spec file itself contains:
make TEST_VERBOSE=1 test

although I would expect something like:

make TEST_VERBOSE=1 test || :

bryteise:
Should be fixed now.

jurobystricky:
Thanks. Works fine now.

PR #256, opened 2018-09-19
Set test failure ignore as part of test string

bryteise:
With the change to how test failures are ignored, made as part of
f9b2da17b19af912aaff83f0205d62434faa6541, add back the ability to
ignore failures in non make and cmake build systems. In this case on
python and perl are currently supported for now.


bryteise:
Resolves #254.

PR #257, opened 2018-09-24
Move license install path

bryteise:


bryteise:
Sorry, pushed before I noticed I didn't merge the first patch already. Just the license file move needs to be reviewed.

PR #258, opened 2018-09-24
Make the main package depend on this build's subpackages

thiagomacieira:
Commit cf9f5e539425834ab72344a3aa6ca0efbac5c10a did that for each of
the sub-packages, but we forgot to do the same for the main package.


bryteise:
+1 after the test case update.

PR #259, opened 2018-09-25
Use the "llvm" package for using Clang

thiagomacieira:
The llvm-dev package is for developing against Clang & LLVM libraries.

LLVM package release 71 has the necessary fixes.


thiagomacieira:
We kinda need this now.

PR #266, opened 2018-10-02
Man pages, and licenses, and locales, should always just be 0644.

ahkok:


phmccarty:
I haven't tested it, but I'm not sure whether `%defattr` works for the locales subpackage, because autospec uses the `%files locales -f NAME.lang` idiom. Is that case confirmed working?

ahkok:
Nope, I was just testing this on `lshw` and it'll make it spew:

```
warning: line 113: second %files
```

ahkok:
I took out the locales. I think we need to do rpm macros to get locales to not have executable permissions.

PR #270, opened 2018-10-03
Add configure_{avx2,avx512} files

thiagomacieira:
Those contain extra auto(re)conf arguments to be passed to AVX2 and
AVX512 builds, instead of configure64.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -225,6 +225,11 @@ configure
   ``./configure`` for an autootools based tarball would result in ``%configure
   --disable-static`` being emitted in the ``.spec``.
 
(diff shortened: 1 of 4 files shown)


thiagomacieira:
This is required for an upcoming change in Mesa, to enable the SWR Gallium driver only for AVX2 builds. See https://bugs.freedesktop.org/show_bug.cgi?id=108135. This is related to clearlinux/distribution#185.

Besides Mesa, Clear only currently has two more configure64 files: curl and NetworkManager. Neither package has AVX2 builds enabled, so they won't be affected.

PR #273, opened 2018-10-04
Don't package scripts as licence files

thiagomacieira:
Just becuase they start with "licence" or "copyright" does not mean
they are licence texts. They can be scripts, as this one in qemu does:
 https://github.com/qemu/ipxe/blob/master/src/util/licence.pl

PS: "licence" is not misspelling. It's the correct spelling for those
of us who use British English (see
https://en.wiktionary.org/wiki/licence#English)

### DIFF
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -24,6 +24,7 @@
 import sys
 import os
 import re
+import hashlib
 import shlex
 import tarball
 import pycurl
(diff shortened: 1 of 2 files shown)


autospec/license.py
```diff
@@ -126,12 +126,13 @@ def scan_for_licenses(srcdir):
                "notice",
                "copyrights",
                "about_bsd.txt"]
-    # look for files that start with copying or licen[cs]e (spelling errors)
-    # or end with licen[cs]e
+    # look for files that start with copying or licen[cs]e (but are
+    # not likely scripts) or end with licen[cs]e
     target_pat = re.compile(r"^((copying)|(licen[cs]e))|(licen[cs]e)$")
+    script_pat = re.compile(r"\.(sh|pl|py|rb)$")
```
bryteise: How do you feel about looking for shebang's to start the file in addition to/in place of this?

thiagomacieira: I feel like it's a great idea. Trying now.

Looking at what files we have packaged as licences and would no longer be so, I can only find one: /usr/share/doc/autogen/agen5_test_license.test

PR #276, opened 2018-10-10
Change the signature files handling.

busykai:
Instead of probing the url with HEAD, download the first available
signature, if any, and pass it to validator.

The concept of "validation from url" is, thus, gone. Downloading the
signature file is no longer part of the validation.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -177,10 +177,6 @@ class Verifier(object):
         print(SEPT)
 
     @staticmethod
-    def download_file(url, destination):
-        return attempt_to_download(url, destination)
-
-    @staticmethod
     def quit():
         print('Critical error quitting...')
         exit(1)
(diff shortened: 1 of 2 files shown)


autospec/pkg_integrity.py
```diff
@@ -210,32 +206,43 @@ def __del__(self):
...
-    curl.setopt(curl.NOBODY, True)
-    curl.setopt(curl.TIMEOUT, 5)
-    curl.setopt(curl.FOLLOWLOCATION, True)
-    curl.perform()
-    http_code = curl.getinfo(pycurl.HTTP_CODE)
-    curl.close()
-    return http_code
+def download_file(url, destination):
+    sign_file = os.path.join(destination, os.path.basename(url))
+    print("Attempting {}".format(sign_file))
```
bryteise: Would probably be more verbose if we are adding a comment for this.

autospec/pkg_integrity.py
```diff
@@ -210,32 +206,43 @@ def __del__(self):
...
+            print(e.args)
+            return None
+        code = curl.getinfo(pycurl.HTTP_CODE)
+        curl.close()
+        if code != 200:
+            os.unlink(sign_file)
+            return None
+        print("Downloaded {}!".format(sign_file))
+        return sign_file
+    return None
```
bryteise: I think this return None is not reachable.

autospec/pkg_integrity.py
```diff
@@ -210,32 +206,43 @@ def __del__(self):
...
+def get_signature_file(package_url, package_path):
     if 'samba.org' in package_url:
-        return package_url + '.asc'
+        return download_file(package_url + '.asc', package_path)
     elif '://pypi.' in package_url[:13]:
-        return package_url + '.asc'
+        return download_file(package_url + '.asc', package_path)
     elif 'mirrors.kernel.org' in package_url:
-        return package_url + '.sig'
+        return download_file(package_url + '.sig', package_path)
```
bryteise: I'm ambivalent about keeping these and just letting any type be tried. I don't see it adding a major slowdown especially as it is only downloaded once.

autospec/pkg_integrity.py
```diff
@@ -318,25 +325,20 @@ def __init__(self, **kwargs):
...
     @staticmethod
     def get_shasum(package_url):
-        data = BytesIO()
-        curl = pycurl.Curl()
-        curl.setopt(curl.URL, package_url)
-        curl.setopt(curl.WRITEFUNCTION, data.write)
-        curl.setopt(curl.FOLLOWLOCATION, True)
-        curl.perform()
-        return data.getvalue().decode('utf-8')
+        for shasum_url in (package_url.replace(".tar.xz", ".sha256sum"),
```
bryteise: Someday we'll try replacing zips and tar.gzs too meh.

bryteise:
Seems fine for me.

autospec/pkg_integrity.py
```diff
@@ -210,32 +206,43 @@ def __del__(self):
...
+            curl.perform()
+        except pycurl.error as e:
+            print(e.args)
+            return None
+        code = curl.getinfo(pycurl.HTTP_CODE)
+        curl.close()
+        if code != 200:
+            os.unlink(sign_file)
+            return None
+        print("Downloaded {}!".format(sign_file))
```
phmccarty: I don't think this message needs the exclamation mark

autospec/pkg_integrity.py
phmccarty: Yeah, I would be happy with printing `Attempting to download ...` instead.

autospec/pkg_integrity.py
```diff
@@ -210,32 +206,43 @@ def __del__(self):
...
-    curl.close()
-    return http_code
+def download_file(url, destination):
+    sign_file = os.path.join(destination, os.path.basename(url))
+    print("Attempting {}".format(sign_file))
+    with open(sign_file, 'wb') as f:
+        curl = pycurl.Curl()
+        curl.setopt(curl.URL, url)
+        curl.setopt(curl.WRITEDATA, f)
+        curl.setopt(curl.FOLLOWLOCATION, True)
```
phmccarty: Just a suggestion: Perhaps set `curl.FAILONERROR` too, so that http codes > 400 will automatically result in an error? Because the previous code didn't do this, it's not a huge deal right now.

autospec/pkg_integrity.py
busykai: Sorry, couple of prints are leftovers.

busykai:
@bryteise, @phmccarty so no issue with the general approach (to getting rid of problematic `HEAD` request), then? i'll finish up the PR and the tests and will ping you again then. thanks for the comments and sorry for `print`s. :)

phmccarty:
@busykai This approach is fine with me.

autospec/pkg_integrity.py
busykai: I'll keep this logic under pretense that it might be restricting the types of signatures that are expected that specific location.

busykai:
Ready for review.

bryteise:
Looks good.

PR #279, opened 2018-10-10
Complement last commit: the main CMake build needs {build,make}_prepend

thiagomacieira:


bryteise:
Sad I missed these when I did the pre, install, post work =/. Thanks for catching these.

PR #281, opened 2018-10-12
autospec/specfiles.py: only add zero-caller flags for x86

clsulliv:
This flag is currently only available for x86, so only append when
autospec is being ran from that architecture.

Signed-off-by: California Sullivan <california.l.sullivan@intel.com>


bryteise:
Alrighty, specfiles.py really need a relook at this point. Thinking of arch considerations as part of this work is also important as I don't want to have this kind of code littered in various functions like avx support.

PR #282, opened 2018-10-12
Add support for subdir option for setup tools

VictorRodriguez:
Since there are some python packages that dont have their setup.py in
the root of the tarball, this patch helps to fix this issue in build and
install section

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -973,6 +973,8 @@ class Specfile(object):
         self.write_prep()
         self.write_lang_c(export_epoch=True)
         self.write_variables()
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -973,7 +973,11 @@ def write_distutils_pattern(self):
         self.write_prep()
         self.write_lang_c(export_epoch=True)
         self.write_variables()
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
         self._write_strip("python2 setup.py build -b py2 " + config.extra_configure)
+        if self.subdir:
```
bryteise: Probably need to do this for the tests below (if tests.config line)

bryteise:
Thanks!

PR #284, opened 2018-10-12
Move the install_prepend after the rm -rf %{buildroot}

thiagomacieira:


bryteise:
That seems fine though I am somewhat unsure when install_prepend should be used to add things to the buildroot.

thiagomacieira:
Yeah, I'm using it for something I shouldn't. It's not something we'd like to do, but it should unblock the LLVM/Mesa mess.

bryteise:
Fair enough.

PR #288, opened 2018-10-18
Allow qmake pattern to use `subdir`.

ahkok:


ahkok:
@bryteise like this?

autospec/specfiles.py
```diff
@@ -1259,17 +1259,26 @@ def write_qmake_pattern(self):
...
         self.write_variables()
 
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
         self._write_strip("%qmake {}{}".format(extra_qmake_args, config.extra_configure))
         self._write_strip("test -r config.log && cat config.log")
         self.write_make_line()
         if config.config_opts['use_avx2']:
+            if self.subdir:
```
bryteise: That'll work though I'd just do the
```
if self.subdir:
                self._write_strip("popd")
```
after the `self.write_make_line()` call directly and not inside the avx2 check. Saves from needing the else. Thoughts?

ahkok: too many meetings today.

ahkok:
""First Try!""

PR #289, opened 2018-10-19
Parse multi-license string from the license server

busykai:
The license server (as it becomes smarter) is able to identify
space-separated multiple licenses, e.g. "Artistic-1.0 GPL-2.0" and
autospec needs to be able to understand the response correctly.


bryteise:
@busykai Seems fine aside from the gripe from the check =P

busykai:
@bryteise cleaned up the scurf. :)

PR #290, opened 2018-10-22
Unlink integrity files when pycurl errors occur

phmccarty:
Setting `curl.FAILONERROR` means that pycurl will raise an error for certain conditions that previously succeeded. Make sure to unlink integrity files for these error cases.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -223,6 +223,7 @@ def download_file(url, destination):
             curl.perform()
         except pycurl.error as e:
             print(e.args)
+            os.unlink(sign_file)
             return None
         code = curl.getinfo(pycurl.HTTP_CODE)
         curl.close()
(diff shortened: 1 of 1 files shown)


autospec/pkg_integrity.py
```diff
@@ -223,6 +223,7 @@ def download_file(url, destination):
             curl.perform()
         except pycurl.error as e:
             print(e.args)
+            os.unlink(sign_file)
```
bryteise: Could we just not put the perform in a try catch? I don't think unlink inside the open block for the same file is very nice =/.

phmccarty: Sure, I will fix...

busykai: That's a good point. We still do this below when checking the HTTP code.

busykai:
Thanks, @phmccarty! Much better now.

PR #292, opened 2018-10-24
Fix cmake srcdir properly.

ahkok:
use `subdir` file when you need this.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -1150,7 +1150,6 @@ class Specfile(object):
 
     def write_cmake_pattern(self):
         """Write cmake pattern to spec file."""
-        self.subdir = "clr-build"
         if self.extra_make == "" or self.extra_cmake == " ":
             self.extra_make = "VERBOSE=1"
         self.write_prep()
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -599,6 +599,9 @@ class Specfile(object):
                 file2 = file.replace("/", "_")
                 self._write_strip("cp " + file + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + "\n")
 
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
         if config.config_opts['32bit']:
             self._write_strip("pushd clr-build32")
             self._write_strip("%make_install32 {} {}".format(self.extra_make_install,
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -1178,11 +1185,12 @@ def write_ruby_pattern(self):
 
     def write_cmake_pattern(self):
         """Write cmake pattern to spec file"""
-        self.subdir = "clr-build"
         if self.extra_make == "" or self.extra_cmake == " ":
             self.extra_make = "VERBOSE=1"
         self.write_prep()
         self.write_lang_c(export_epoch=True)
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
```
bryteise: Hrm okay this makes sense to me but it also would make sense to me to replace the cmake_srcdir with subdir in the %cmake macro call below. I do think we should only need one of these however though. Thoughts?

phmccarty: My understanding is that `cmake_srcdir` is relative to the clr-build directory (which is within `subdir` if set), so I think the code you're referring to is correct as-is

ahkok:
We need to add migration code so old packages are converted automatically.

ahkok:
Fixed (hopefully) avx512 builds now, too

ahkok:
Once more, with feeling.

autospec/specfiles.py
```diff
@@ -653,9 +657,14 @@ def write_cmake_install(self):
             self._write_strip("%s_avx2 %s || :\n" % (self.install_macro, self.extra_make_install))
             self._write_strip("popd")
 
+
```
phmccarty: flake8 is complaining about this blank line...

autospec/specfiles.py
```diff
@@ -1241,6 +1253,10 @@ def write_cmake_pattern(self):
 
         self._write_strip("\n")
         self.write_check()
+
+        if self.subdir:
+            self._write_strip("popd")
```
phmccarty: The `popd` needs to be added before the `self.write_check()` two lines above, because that function will write a `%check` section if needed.

bryteise: So looking at the code, it adds ```cd clr-build; make test``` as its test code, in this case I think it needs the popd to be after the write_check.

phmccarty: But the test code lives in a `%check` section, right? So that code is run in a different script...

bryteise: Ugh right, of course. That's a bit too ugly to support =/.

bryteise: Updated.

phmccarty:
I rebased the branch again.

phmccarty:
The changes were working for me before, but not anymore... I'm debugging it.

bryteise:
Okay this all makes sense to me. Could you validate that it can build subdir setting and not setting cmake projects with this change and then I'll merge it?

PR #293, opened 2018-10-24
commitmessage: omit merge commits

phmccarty:
For the `git log` and `git shortlog` output, omit merge commits, since it's generally more useful for autospec to record only the summaries of what actually changed in the code itself.


bryteise:
Merging as the issues with linting need to be fixed outside of this PR.

PR #294, opened 2018-10-24
Fix linter (flake8) errors.

busykai:


phmccarty:
Looks like flake8 is still unhappy. But there are fewer errors than before.

busykai:
@phmccarty yep, I needed a break from it. :)

One items that needs a discussion here is the following: flake8 complains about both "line break before binary operator" (W503) and "line break after binary operator" (W504). Since it's unrealistic and unnecessary to have all the binary operators in one line, we need to choose one. I, personally, prefer break before binary op, so I fixed it that way. I.e. currently:

Yes:
``` python
a = b
    + c
```
No:
```python
a = b +
    c
```

PR #298, opened 2018-11-01
Check license server http return code

alexjch:
This change verifies that the HTTP returning code from license server is
a 200 otherwise any code returned by the server will be accepted as a
successful response and the returned error page will be interpreted as a
valid license hash.

Additonally this change narrows the exception emitted by pycurl perform
method from Exception to pycurl.error.

Signed-off-by: Alex Jaramillo <alex.v.jaramillo@intel.com>

### DIFF
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -94,7 +94,8 @@ def license_from_copying_hash(copying, srcdir):
         c.setopt(c.URL, config.license_fetch)
         c.setopt(c.WRITEDATA, buffer)
         c.setopt(c.POSTFIELDS, data)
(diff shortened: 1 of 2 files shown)


autospec/license.py
```diff
@@ -97,13 +97,17 @@ def license_from_copying_hash(copying, srcdir):
         c.setopt(c.FOLLOWLOCATION, 1)
```
phmccarty: I recommend setting the `FAILONERROR` option here, so that "fatal" HTTP responses will also raise an exception.

See the recent changes to pkg_integrity.py for an example.

PR #299, opened 2018-11-02
Display URL attempted for signature file download

bryteise:
When displaying an error message for signature file downloads, display
the URL that was attempted as part of the error message as it is more
useful than the error code.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -244,8 +244,8 @@ def get_signature_file(package_url, package_path):
                 return sign_file
             elif os.path.exists(dest):
                 os.unlink(dest)
-        except pycurl.error as e:
-            print(e.args)
+        except pycurl.error:
+            print(f"Download of signature file {url} failed")
             if os.path.exists(dest):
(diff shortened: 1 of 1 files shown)


autospec/pkg_integrity.py
```diff
@@ -244,8 +244,8 @@ def get_signature_file(package_url, package_path):
                 return sign_file
             elif os.path.exists(dest):
                 os.unlink(dest)
-        except pycurl.error as e:
-            print(e.args)
+        except pycurl.error:
+            print(f"Download of signature file {url} failed")
```
phmccarty: Do we care to print the HTTP status code here?

bryteise: I didn't think it really mattered as folks who are confused as to why a URL didn't work probably try it anyway, most of the time failure is expected as it is just guessed URLs being tried.

phmccarty: Okay, just checking, since printing `e.args` did report the status code :) I don't have strong opinions either way.

PR #303, opened 2018-11-06
Don't insert -m64 here - it breaks 32bit builds.

ahkok:


miguelinux:
There is a 32bit section

https://github.com/clearlinux/autospec/blob/master/autospec/specfiles.py#L406

Is that section broken?

ahkok:
this is the section for `conservative_flags` which gets added on top of the 64-or-32 specific flags. Therefore, in the rare case a package has 32bit builds and `conservative_flags` gets set, it breaks.

PR #304, opened 2018-11-06
Test cleanups

phmccarty:
- To make the import ordering more sensible and remain pep8 compliant, rename `autospec/test.py` to autospec/check.py`.
- Fix `make test_specfile` and `make test_test`
- Fix `autospec/count.py` tests due to the module renaming results in reordering of test execution.

@bryteise is looking into the remaining test failures in pkg_integrity.


tests/test_specfile.py
```diff
@@ -13,6 +13,7 @@ def setUpClass(self):
     def setUp(self):
         url = "http://www.testpkg.com/testpkg/pkg-1.0.tar.gz"
         self.specfile = specfiles.Specfile(url, '1.0', 'pkg', '2')
+        specfiles.config.config_opts['dev_requires_extras'] = False
```
bryteise: This is fine, though it exposes similar timing problems because it doesn't follow a backup and restore once finished pattern. We don't do that hardly anywhere though so I think it is more symptomatic of the problems autospec has with global state mutation so more reasons to refactor.

tests/test_pkg_integrity.py
```diff
@@ -94,7 +91,7 @@ def _mock_fetch_shasum(url):
     @mock.patch('pkg_integrity.GnomeOrgVerifier.fetch_shasum', _mock_fetch_shasum)
     def test_gnome_org(self):
         result = self.run_test_for_domain(pkg_integrity.GnomeOrgVerifier, GNOME_SHA256_PKG)
-        self.assertTrue(result)        
+        self.assertTrue(result)
 
 
 @mock.patch('pkg_integrity.download_file', mock_download_file)
```
phmccarty: Is this `@mock.patch` (and others) still needed, since you are removing the download_file calls in tests?

bryteise: Yes, the download_file function is called by most of the tests under the pkg_integrity.check() function.

phmccarty: Oh, right, thanks for clarifying that.

PR #307, opened 2018-11-12
Make ChangeLog files read-writeable

thiagomacieira:
Some sources ship them read-only, which resulted in errors down the
line.


bryteise:
Oh fun, thanks!

PR #308, opened 2018-11-19
Implement mock root.log error detection

phmccarty:
Occasionally, packages will add new dependencies that are not yet available in the repos, or autospec may add an invalid build dependency for the next build round, etc.

To report these error conditions in a more meaningful way, scan the root.log after a build whenever mock exits with a non-zero code. If mock logged a missing dependency (or multiple missing dependencies), report that information to the user. Otherwise, consider the root.log to be "clean", and continue by scanning the build.log to determine how to proceed.

Also, add a new unit test for the root.log parsing function.


autospec/build.py
```diff
@@ -158,6 +158,28 @@ def failed_pattern(line, pattern, verbose, buildtool=None):
             warned_about.add(s)
 
 
+def parse_buildroot_log(filename, returncode):
+    """Handle buildroot log contents."""
+    if returncode == 0:
+        return
+    else:
```
bryteise: May as well save the indent and skip the else clause here since you return otherwise.

phmccarty: Okay, will fix shortly, thanks

PR #316, opened 2018-12-03
Implement unified download interface

phmccarty:
The download code for licenses, tarballs, and signature files are
similar enough to warrant consolidation into a unified interface, so add
a new function do_curl() that satisfies the needs of all the callers.

Note that I made one change that noticeably changes the output,
hopefully to reduce confusion: The signature download loop no longer
prints error messages, since the end user should not need to care at all
unless a signature cannot be downloaded and verification is required. If
the latter case is true, autospec will still exit and print the fatal
error, so the issue should be straightforward to diagnose after seeing
that error.

Also, a couple of bugs are fixed as a result of using the new function:

  - The FAILONERROR pycurl option is now set for tarball downloads. This
    prevents certain undesirable side-effects like downloading 404
    response pages.

  - Responses from FTP servers are now handled better, since a more
    common "success" code from FTP servers is 226, not 200.

### DIFF
--- a/autospec/download.py
+++ b/autospec/download.py
@@ -1,0 +1,75 @@
+#!/usr/bin/true
+#
+# download.py - part of autospec
+# Copyright (C) 2018 Intel Corporation
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from io import BytesIO
+import os
+import sys
+
+import pycurl
+from util import print_fatal
+
+
+def do_curl(url, dest=None, post=None, is_fatal=False):
+    """
+    Perform a curl operation for `url`.
+
+    If `post` is set, a POST is performed for `url` with fields taken from the
+    specified value. Otherwise a GET is performed for `url`. If `dest` is set,
+    the curl response (if successful) is written to the specified path and the
+    path is returned. Otherwise a successful response is returned as a BytesIO
+    object. If `is_fatal` is `True` (`False` is the default), a GET failure,
+    POST failure, or a failure to write to the path specified for `dest`
+    results in the program exiting with an error. Otherwise, `None` is returned
+    for any of those error conditions.
+    """
+    c = pycurl.Curl()
+    c.setopt(c.URL, url)
+    if post:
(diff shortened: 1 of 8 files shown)


autospec/download.py
```diff
@@ -0,0 +1,71 @@
...
+    finally:
+        c.close()
+
+    # write to dest if specified
+    if dest:
+        try:
+            with open(dest, 'wb') as fp:
+                fp.write(buf.getvalue())
+        except Exception as e:
+            if os.path.exists(dest):
```
bryteise: I think I'm inclined to leave cleanup of the file to open rather than handle unlinks in the exception handler.

bryteise:
Looks good overall to me.

autospec/download.py
phmccarty: If an exception is raised, I don't see the file unlinked though... I left it out originally

bryteise: Hrm well my worry is that the file can't be opened for writing due to it being opened by another process and this will cause another exception. Maybe an exception for IOError that does the unlink and a generic exception that doesn't? unlink is annoying in that there isn't an option for it to ignore failures.

bryteise: Perhaps don't catch the generic exception even.

phmccarty: Okay, I will just handle `IOError`

phmccarty: I made this change in the latest commit

phmccarty:
Thanks, will fix up the unit tests now...

phmccarty:
@bryteise Existing unit tests are fixed now. I am working to add tests specifically for `do_curl()`. Do you want those ready before merge?

bryteise:
@phmccarty I'd prefer tests for do_curl come with unless this is particularly urgent.

phmccarty:
@bryteise I have added the new unit tests. Ready to re-review.

tests/test_download.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @patch('download.pycurl.Curl')
+    def test_download_get_success_no_dest(self, test_curl):
+        """
+        Test successful GET request when dest is not set.
+        """
+        instance = init_curl_instance(test_curl)
+        instance.setopt.side_effect = test_opts
+        data = download.do_curl("foo")
+        print(test_curl.mock_calls)
```
bryteise: Probably an extra print here.

tests/test_license.py
```diff
@@ -167,7 +168,7 @@ def getvalue(_):
                 return 'GPL-3.0'.encode('utf-8')
 
         # set the mocks
-        license.BytesIO = MockBytesIO
+        download.BytesIO = MockBytesIO
```
bryteise: Probably need to reset this I would guess as I see further down license.BytesIO is reset instead.

bryteise:
@phmccarty Looks good outside of what I think are minor issues. Very nice work!

tests/test_download.py
phmccarty: Yep, this was a leftover from debugging the test. I removed it.

tests/test_license.py
phmccarty: Good catch. I fixed the line that resets it below.

PR #317, opened 2018-12-04
Set MAKEFLAGS for Python3 distutils

thiagomacieira:
Because the Python script runs make, we don't get the chance to pass
the -j option directly.


bryteise:
Seems safe enough and obviously I'd like to have python builds compile C in parallel.

PR #318, opened 2018-12-04
Add a few more MAKEFLAGS to Python distutils

thiagomacieira:
For Python2 and for the %install step. Some packages build during
installation...


bryteise:
Ah install too, thanks!

PR #319, opened 2018-12-06
Add support for generic sha256 verification

thiagomacieira:
So long as get_hash_url returns a URL, we'll try to use it

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -348,9 +348,9 @@ class MD5Verifier(Verifier):
         return md5_digest == self.md5_digest
 
 
-# gnome.org Verifier
-class GnomeOrgVerifier(ShaSumVerifier):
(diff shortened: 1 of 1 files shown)


phmccarty:
I would like to get my PR integrated (hopefully by tomorrow) before tackling this issue.

phmccarty:
I don't see any obvious problems with this approach though.

thiagomacieira:
We can wait. I'll have a few more autospecs along the day and will find/fix issues with this.

bryteise:
Okay, now we can take another look at this one.

phmccarty:
@thiagomacieira So, I think the intent here is for `GenericDownloadVerifier` to first construct several hash URLs and try to download each in turn. Right now, it is still using `get_hash_url`, which only understands how to handle download.gnome.org and download.qt.io domains.

I can take over the work on this if you prefer.

PR #322, opened 2018-12-19
Set CMAKE_INSTALL_LIBDIR for 32 bits builds

cmarcelo:
Besides the LIB_INSTALL_DIR the %cmake macro now sets the
CMAKE_INSTALL_LIBDIR to /usr/lib64, since some things use either one
or other.  So we need to override it in 32 bits builds to avoid it
mistakenly installing 32 files into lib64.


bryteise:
Thanks!

PR #324, opened 2019-01-02
specfiles: Avoid unnecessary whitespace in -m32 writes

ikeyd:
With newer CMake versions, having unsanitized " -m32" LDFLAGS with the
whitespace will violate policy CMP0004 causing build failures in SDL2.
Thus we modify the .spec file to only set "-m32" if the string is empty,
otherwise include " -m32" and the existing variable.

Signed-off-by: Ikey Doherty <ikey.doherty@intel.com>


ikeyd:
ta

PR #326, opened 2019-01-10
Add support for lzip-compressed tarballs (.tar.lz)

thiagomacieira:
Our GNU tar does recognise it and will run /usr/bin/lzip on its
own. Only that it needs to be installed in the chroot for that to
work.

Homepage: https://www.nongnu.org/lzip/
See also: https://www.nongnu.org/lzip/xz_inadequate.html

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -67,6 +67,9 @@ def check_or_get_file(upstream_url, tarfile):
 
 def build_untar(tarball_path):
     """Determine extract command and tarball prefix from tar -tf output."""
+    if tarball_path.lower().endswith('.lz'):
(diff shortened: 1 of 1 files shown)


thiagomacieira:
How?

Anyway, this is so uncommon that I didn't think that "polluting" our build environments with lzip was worth it.

bryteise:
That's a good point though zip is pretty rare too and I don't think the tooling is adding much. Unfortunately the how is not super obvious as it requires some infrastructure change. I'll poke devops folks and see which they would prefer and get back to you.

bryteise:
Okay, the build root was updated so this change isn't needed.

PR #327, opened 2019-01-10
Honor requires_ban for subpackages of base

phmccarty:
In case a package should not require one or more subpackages that are declared by default, skip the ones listed in requires_ban.


bryteise:
Thanks!

PR #328, opened 2019-01-16
Don't parse PDF files for descriptions

seanvk:
Signed-off-by: Sean V Kelley <sean.v.kelley@linux.intel.com>


autospec/specdescription.py
```diff
@@ -242,6 +242,8 @@ def scan_for_description(package, dirn):
     """Scan the project directory for things we can use to guess a description and summary."""
     for dirpath, _, files in os.walk(dirn):
         for name in files:
+            if name.lower().endswith(".pdf"):
+                break
```
bryteise: I think this should be a continue rather than a break otherwise looks good.

seanvk: Ah, yeah I agree.

PR #332, opened 2019-01-31
Add new failed pattern handler

bryteise:


autospec/config.py
```diff
@@ -265,6 +266,7 @@
     (r"ImportError:..*: No module named ([a-zA-Z0-9\-\._]+)", 0, 'pypi'),
     (r"ImportError: No module named ([a-zA-Z0-9\-\._]+)", 0, 'pypi'),
     (r"ImportError: No module named '([a-zA-Z0-9\-\._]+)'", 0, 'pypi'),
+    (r"ImportError: No module named '([a-zA-Z0-9\-\._]+)'", 0, 'pypi'),
```
phmccarty: I don't see a difference between this pattern and the previous one... Is this intentional?

bryteise: Nope, that's a mistake, thanks

bryteise:
Updated.

PR #333, opened 2019-02-01
Support decompressing ZIP files with comments

lpereira:
While trying to determine the directory name when extracting a ZIP
file, autospec do not consider the (optional) ZIP archive comment,
which is printed to the console while printing the archive contents
with "-l".

By calling unzip with "-q", the archive comment is not printed, and
autospec is happy again.


bryteise:
Ah, we have so few zip file repos I missed this. Thanks for the fix!

PR #335, opened 2019-02-15
Save buildreq_cache on failure

bwarden:
In cases where autospec spends multiple rounds resolving dependencies,
before ultimately failing, it would be nice to save the buildreq_cache.
This could save a lot of time while attempting to whittle away actual
build problems.


bryteise:
Yea, that's a nice win. Thanks!

PR #337, opened 2019-02-25
cmake: Ignore entries inside comments

cmarcelo:
Change the regexp to match only non-comment lines.  Found when I
figured waffle was depending on "pkgconfig(foo)".  Added a test for
that case.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -304,8 +304,8 @@ def rakefile(filename):
 
 def parse_cmake(filename):
     """Scan a .cmake or CMakeLists.txt file for what's it's actually looking for."""
-    findpackage = re.compile(r"find_package\((\w+)\b.*\)", re.I)
-    pkgconfig = re.compile(r"pkg_check_modules\s*\(\w+ (.*)\)", re.I)
+    findpackage = re.compile(r"^[^#]*find_package\((\w+)\b.*\)", re.I)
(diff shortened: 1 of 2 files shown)


autospec/buildreq.py
```diff
@@ -304,8 +304,8 @@ def rakefile(filename):
 
 def parse_cmake(filename):
     """Scan a .cmake or CMakeLists.txt file for what's it's actually looking for."""
-    findpackage = re.compile(r"find_package\((\w+)\b.*\)", re.I)
-    pkgconfig = re.compile(r"pkg_check_modules\s*\(\w+ (.*)\)", re.I)
+    findpackage = re.compile(r"^[^#]*find_package\((\w+)\b.*\)", re.I)
```
thiagomacieira: Should we catch spaces before the `#` ?

cmarcelo: Yes.

cmarcelo: Actually this regexp is already correct. It accepts any prefix that doesn't contain the comment character.

cmarcelo: So it already rejects `    # find_package...`, because it contains a comment char in the prefix.

thiagomacieira:
Please add test:
```cmake
    #find_package(foo)
```
(with the spaces before the `#`)

cmarcelo:
New test added.

bryteise:
Looks good, thanks!

Issue #341, opened 2019-02-28
Do not create/update ChangeLog if giturl setting is defined


phmccarty:
Because autospec adds `git shortlog` output to commits when the `giturl` setting is defined in options.conf, creating or updating the ChangeLog is usually not very useful in this situation. If one wants more detail than that provided by the shortlog, output from the appropriate `git log`, `git log -p`, or other commands can be examined instead.

bryteise:
I was thinking about completely ignoring the changelog eventually but I'm going to do a first pass with just ignoring on giturl being set first and we'll go from there.

phmccarty:
Resolved with #342

PR #342, opened 2019-02-28
Ignore the ChangeLog when the giturl is available

bryteise:
Often the ChangeLog is a poor duplicate of the git log for a given
release and introduces unnecessary noise into package updates. To
avoid this only use the ChangeLog if the giturl is undefined.


bryteise:
Fixes #341

PR #344, opened 2019-03-01
Set GOPATH only if no go.mod file exists

bwarden:
This detects _any_ go.mod within the tarball, and if found, will _not_ write out the `export GOPATH` statement in the spec file.


bryteise:
That'll do it, thanks!

PR #346, opened 2019-03-05
Slightly cleanup lookup for go.mod

bwarden:
Don't need the `for` loop, although it probably does the same thing under the hood.


bryteise:
Ah that's nicer to read, thanks!

PR #347, opened 2019-03-06
Detect and decode license text codepages

bwarden:
License text could be in different codepages. Attempt to autodetect the
correct codepage, with some logic to catch commonly-seen escapes. Decode
the bytestream, according to that codepage, into a native Python string.
This will usually result in sending proper unicode to the license
server.


bryteise:
Looks good, thanks!

PR #348, opened 2019-03-08
Fix typo for KF5Kipi mapping

phmccarty:
The base package name is 'libkipi', not 'kipi'.


bryteise:
Oops, thanks for the fix!

PR #351, opened 2019-03-14
Avoid crashing on binary files that may look like a license file.

ahkok:
Test case: libreoffice. It has several files with the word "license"
and "copying" in the filename, which are ZIP files and thus not
any valid encoding. These files need to be skipped entirely.

Forcing an encoding here just propagates the problem as things like
ASCII and utf-8 will just move the crash down a few lines.

### DIFF
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -95,6 +95,10 @@ def license_from_copying_hash(copying, srcdir):
             license_charset = 'ISO-8859-13'
         elif b'\xd2' in data and b'\xd3' in data:
             license_charset = 'mac_roman'
+    if not license_charset:
+        """This is not a text file"""
+        return
+
     data = data.decode(license_charset)
 
     if config.license_fetch:


ahkok:
@bwarden

bwarden:
We could use `python-magic` to make a better first pass.

bwarden:
lgtm, though.

ahkok:
That's totally reasonable, but I think we need to fall through at some point anyway.

autospec/license.py
```diff
@@ -95,6 +95,10 @@ def license_from_copying_hash(copying, srcdir):
             license_charset = 'ISO-8859-13'
         elif b'\xd2' in data and b'\xd3' in data:
             license_charset = 'mac_roman'
+    if not license_charset:
+        """This is not a text file"""
```
bryteise: I'm just going to fix up this comment, looks good otherwise.

PR #352, opened 2019-03-19
Update README.rst

tripzero:
Add example series line with patch arguments


bryteise:
That's a very helpful bit of documentation, thanks!

PR #355, opened 2019-03-27
Support using the auto-detected name

bwarden:
Rework to absorb a lot of the work we were doing in makefiles so that we
can avoid setting the package name manually in 'make autospecnew'.

Do not merge yet.


bwarden:
You'll also need to modify Makefile.toplevel to:

1. Remove the check for `$(NAME)`
1. Remove `$(MAKE) clone_$(NAME)` and the `mkdir`/`git` block
1. Don't write the Makefile (autospec will do that instead)
1. Remove the `--target` and `--name` parameters from autospec
1. Add the `--package-dir packages/` param _to_ autospec
1. Remove the `else` block for `$(NAME) already exists...`
1. Add `git_pull` to `autospec.conf` if you use a different git URI for pull than push.

bwarden:
The `link-new-rpms` and `checkblacklist` actions will also need to be modified/replaced at some point.

bwarden:
@phmccarty, @bryteise Could I have your review, please?

autospec/tarball.py
```diff
@@ -561,6 +570,11 @@ def process(url_arg, name_arg, ver_arg, target, archives_arg, filemanager):
     # name is created by adding ".gcov" to the package name (if a gcov file
     # exists)
     set_gcov()
+    # build the target path, if necessary
+    if not target and package_dir:
+        target = os.path.join(package_dir, name)
+    # create the path, by cloning existing repo or manually, as necessary
+    call("make clone_%s" % name, check=False, stderr=subprocess.DEVNULL)
```
phmccarty: I think autospec should also try to clone the repo with git if the `make clone_PKG` command fails. And if that second clone attempt fails, then autospec will create a new git repo locally, which I believe it already does.

phmccarty: I think autospec should also try to clone the repo with git if the make clone_PKG command fails, since `make clone_PKG` is only defined for current/latest list of packages; packages that formerly existed in the distro may have package repos available, and we want to reuse those.

Then if that second clone attempt fails, autospec can create a new git repo locally, and I believe that's already implemented.

phmccarty: Alternatively, we could run `git clone ...` in Makefile.toplevel (ignoring errors) prior to running autospec, and remove this line.

bwarden: We'd need `config.git_pull_uri` in order to call git clone directly, and that's populated in `autospec.package()` right _after_ `tarball.process()`

And of course toplevel git clone will only pick up what's in `packages`, and might take a while, too. I'm not sure what's the right answer here.

phmccarty: Oh right, I was confusing the chicken with the egg again :-P

Autospec is detecting the package name, so we cannot run a `git clone` before running `make autospecnew`, since the package name is not yet known. Hmm.

phmccarty: Could we do something like the following instead?

```
if package_dir:
    call("git -C {} clone {}".format(package_dir, config.git_pull_uri % {'NAME': tarball.name}))
```

`git_pull_uri` would need to be defined in autospec.conf, of course.

bwarden: But you still don't have config.git_pull_uri at this point.

bwarden:
Remaining work, in case someone else has time to pick up the torch:

- refactor config to read autospec.conf first, then go back and read package configuration files *after* deciding what the package name and path are.
- tackle creation of rpms subdir and hard-linking resulting RPMs. This is the last ${NAME} dependency in the common makefile.

phmccarty:
@bryteise Is this feature still in scope for autospec?

I think it would be nice to autodetect the package name, but due to the extra complexity introduced with the interaction with the common tooling, we were somewhat undecided about what needed to change in that tooling, and autospec, or both, to accommodate...

bryteise:
Hrm I'm a little unsure about all the bits that need to be reworked.

Because `make autospecnew` is probably the way an autodetected name would be made use of, getting the common tooling as much out of the way as possible of autospec doing what needs to be would be sensible. We could also we add a --get-package-name type option for autospec and then common tooling gets the name from that and reruns autospec with the full options like now under the hood.

phmccarty:
I agree that all the name detection logic should be handled by autospec. The idea of adding a `--get-package-name` flag to autospec to replace the `NAME` value for `make autospecnew` also sounds reasonable...

phmccarty:
There was also some discussion at the time about autospec creating the package Makefile (instead of the common tooling) if it doesn't exist, so that consideration could be revisited as well.

PR #358, opened 2019-04-01
[WIP] update test packageurl list

phmccarty:
This PR refreshes the packageurls list used for tarball name/version/giturl tests. Many test failures are expected until we determine what next steps to take.

(See the travis failure log for details)


phmccarty:
One category of failures is for compat packages that are misnamed. Those are not autospec issues; we just need to fix those packages :-)

phmccarty:
The package name/version autodetection code first needs an overhaul, so I don't think it's worth keeping this PR open for reference. Once that code is in better shape, then we can revisit the packageurl list.

PR #359, opened 2019-04-03
Fix excludes v extras.

busykai:
When adding a file to extras (as well as excludes, setuid etc.), the
files are added to the exclude list dynamically. Instead, load all known
files upfront.

Also, process file-configured subpackages (-extras, custom-extras etc.)
first and only then process "built-in" (pattern-based) packages (such as
-lib64, -lib32, -autostart etc.)

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -713,6 +713,7 @@ def parse_config_files(path, bump, filemanager, version):
     for extra in content:
         print("extras for  : %s." % extra)
     filemanager.extras += content
+    filemanager.excludes += content
 
     for fname in os.listdir(path):
         if not re.search('.+_extras$', fname) or fname == "dev_extras":
@@ -731,16 +732,19 @@ def parse_config_files(path, bump, filemanager, version):
                 continue
         print(f"{name}-extras for {content['files']}")
         filemanager.custom_extras[f"{name}-extras"] = content
+        filemanager.excludes += content['files']
 
     content = read_conf_file(os.path.join(path, "dev_extras"))
     for extra in content:
         print("dev for     : %s." % extra)
     filemanager.dev_extras += content
+    filemanager.excludes += content
 
     content = read_conf_file(os.path.join(path, "setuid"))
     for suid in content:
         print("setuid for  : %s." % suid)
     filemanager.setuid += content
+    filemanager.excludes += content
 
     content = read_conf_file(os.path.join(path, "attrs"))
     for line in content:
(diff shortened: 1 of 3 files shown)


busykai:
Aside from generally proper way to avoid multiple inclusions of a file multiple times to a subpackage, this PR solves the following particular issue:

Create a custom_extras and include `/usr/lib/systemd/system/multi-user.target.wants/systemd-networkd.service` to it. autospec also includes it to "built-in" `autostart` subpackage.

Current behavior: the file will appear in both `custom-extras` and `autostart` subpackage.

With this PR: the file will appear in `custom-extras` and will be excluded from `autostart` subpackage.

bryteise:
So the original behavior was intentional though I didn't really like it. This changes makes me happier. @phmccarty MCA should be fine with this right? I am guess we could see some files being dropped as things get autospec rerun on them.

phmccarty:
@bryteise I don't think MCA will complain about this type of change, since it is able to map files back to the srpms after fully expanding the bundle content to files.

@busykai Regarding subpackage file overlap, I thought that there were certain use cases for it, as @bryteise alludes to? However, I'm fine with eliminating the overlap in general, especially when considering the case in question (i.e `autostart` and `custom-extras`).

busykai:
@phmccarty, i don't know about any use cases for that. certainly, if both packages are in the mesh, find duplicates script will be as unhappy as @bryteise, wouldn't it?

phmccarty:
@busykai No, subpackages that are from the same package can install duplicate files. The file duplicate/conflict checks only apply when subpackages are from different packages entirely.

busykai:
thanks, @phmccarty. i did a quick search on all packages as in
```bash
find . -type f -name \*.spec  -exec grep '^\/' {} + | sort | uniq -c | awk '{ if ($1 > 1) print $0 }'
```
it does show some occurrences of double inclusions in subpackages. This search will not catch the wildcards, but autospec-managed specs do not have them anyways.

The only case worth mentioning that I found is that of `autostart` and `extra` subpackages. Some of the `....target.wants/` autostart files are include both in `autostart` and `extra`. Which is not really necessary.

bryteise:
What I want to do here is just check the duplicate entries and ensure the right duplicate is removed. After that I think we have reached an agreement that this is the desired behavior even though there may in theory be a valid use case. Thoughts @phmccarty @busykai ?

busykai:
@bryteise, do you mean in the potentially impacted packages?

this PR actually establishes clear priorities of what is included where. first, explicitly defined extras will be honored, then autospec's implicit packages (like `lib`, `autostart`). in this light, some of explicit entries of systemd `.wants` links should be removed from `extras` and the bundling should be reviewed (just in case someone was including extras for autostart). this is the extent of the impact i found.

busykai:
This can be merged now. I removed all occurrences of double inclusions from the existing packages.

bryteise:
Awesome, thanks!

PR #361, opened 2019-04-05
Add some more patterns

fenrus75:
detect a few more meson patterns

add a staticdev subpackage to host .a files

make -dev packages require: the main package
so that installing -dev gives you a functional component


bryteise:
@fenrus75 I'm a little unsure about having -dev pull in the main subpackage because of the potential to inflate devpkg and dev bundle sizes but we can reassess later. I'll just fix up the formatting complaint and push the change.

PR #363, opened 2019-04-09
Allow unpacking archives in the main tarball's dir

bwarden:
Some software packages break distribution into a core tarball then
additional tarballs with extra features or content. LibreOffice is one
such example. The additional tarballs provide files in the same
directory tree. The most efficient way to unpack all of these with the
current archive functionality is to place them at the same base dir, ./

At two points during autospec, the code to handle archives attempts to
either move or copy the contents of archive tarballs to the final
directory. If ./ is the unpacking directory, the files are already in
the same place, so mv and cp will fail, complaining that src and dst are
the same.

Add '-n' (no clobber) to these mv and cp commands, so that they simply
do nothing when the files are already in the correct final destination.

**Currently testing this with a build of LibreOffice. Do not merge until this has been tested thoroughly.**

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -395,12 +395,16 @@ class Specfile(object):
                                               self.source_index[archive]))
 
         for archive, destination in zip(self.sources["archive"], self.sources["destination"]):
-            self._write_strip("mkdir -p {}"
-                              .format(destination))
-            self._write_strip("cp -rn %{{_topdir}}/BUILD/{0}/* %{{_topdir}}/BUILD/{1}/{2}"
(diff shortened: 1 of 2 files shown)


bwarden:
This at least solved the problem with LibreOffice. I'm open to review/commentary about whether it's likely to break other things.

phmccarty:
@bwarden Thanks, I think this approach will be better. Does LibreOffice build okay with these new revisions?

bwarden:
LO is failing due to an unrelated problem, so I'm not certain yet.

bwarden:
Yes, looks like libreoffice builds fine with this change.

PR #364, opened 2019-04-09
CMake support: Don't make copies of sources for AVX2/AVX512/32-bit builds

thiagomacieira:
The CMake pattern reuses the same sources by doing different
out-of-tree builds. The copied sources are not useful.


bryteise:
Tests pass after rebasing.

PR #367, opened 2019-04-10
Add phpize package support.

ahkok:
This relies on the convention that `phpize` reads a `config.m4` file and from that finds all the needed PHP stuff.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -632,6 +632,8 @@ def scan_for_configure(dirn):
         add_buildreq("buildreq-scons")
     elif buildpattern.default_pattern == "R":
         add_buildreq("buildreq-R")
+    elif buildpattern.default_pattern == "phpize":
+        add_buildreq("buildreq-php")
 
     count = 0
     for dirpath, _, files in os.walk(dirn):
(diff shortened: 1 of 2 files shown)


ahkok:
BTW the `config.m4` method is garbage, we really should check and see if it contains something like `PHP_ARG_WITH` or something similar.

autospec/buildreq.py
```diff
@@ -632,6 +632,9 @@ def scan_for_configure(dirn):
         add_buildreq("buildreq-scons")
     elif buildpattern.default_pattern == "R":
         add_buildreq("buildreq-R")
+    elif buildpattern.default_pattern == "php":
+        add_buildreq("php")
```
bryteise: Do you want to add a buildreq-php package for these?

ahkok: I think so, because I currently see that it already does need:
```
php
php-dev
re2c
```

ahkok:
Reduced the build_pattern strength to `1` since we want this to be a last resort.

ahkok:
Using `buildreq-php` now, since I'm pushing that to koji as well.

ahkok:
Should we name the build pattern `phpize` instead?

bryteise:
I'd be okay using phpize though it mostly depends on if other php specific build systems are in use.

ahkok:
Renamed to phpize. I think this is probably good to go - I'm going to start using this to add some popular php mods.

PR #368, opened 2019-04-14
Honor requires_ban in subpackages as well.

busykai:
The top-level package does honor a subpackage ban in requires_ban.
However, autodependencies do not honor it. Unless they do, it's
meaningless to control subpackages via requires_ban. For example, if I
want to ban pkg-services, top-level package will not require it, but the
content will still be installed since pkg requires pkg-bin, which, in
turn, will require pkg-services. With this change, it will not.

None of the existing packages is impacted by this change.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -234,6 +234,12 @@ class Specfile(object):
         deps["libexec"] = ["config", "license"]
         deps["lib32"] = ["data", "license"]
         deps["python"] = ["python3"]
+        for pkg, pkg_reqs in deps.items():
+            for pkg_req in pkg_reqs:
+                if "{}-{}".format(self.name, pkg_req) in buildreq.banned_requires:
+                    pkg_reqs.remove(pkg_req)
+                    continue
+
         if config.config_opts['dev_requires_extras']:
             deps["dev"].append("extras")
         for k, v in self.custom_extras.items():


bryteise:
Seems fine, I'm probably just not thinking well but why would a subpackage have any odd dependencies? It should just be other subpackage components right?

busykai:
@bryteise, we have auto dependencies in packages. in an odd case where we need to not have them, it turns out to be impossible to control. like openssh where we the server and the client part are produced from a single autospec-managed package and we do not want `-services` subpackage to be installed with the top-level subpackage.

bryteise:
Ah okay, that use case is interesting. Usually that would only be an issue when something else is pulling in the top level package though because the bundle definition file would just specify contents at the subpackage level if it wanted to avoid pulling in the default set of dependencies. I guess this is probably just a generic case of the autostart config option though the reason we made it specific is that we want to have default dependencies be something that folks won't need to think about. Enabling this kind of blacklisting makes me think we might be working around another kind of issue. The openssh example for instance, perhaps we want to move the service files and the enabling symlinks into a custom extras that the openssh-server bundle uses. I'd like to think about use cases more perhaps.

busykai:
> perhaps we want to move the service files and the enabling symlinks into a custom extras that the openssh-server bundle uses.

well, this is exactly what led to this PR. i tried to package the server pieces with `server_extras`, like this:

```
files = [ '/usr/bin/sshd',
          '/usr/libexec/sftp-server' ]
requires = [ 'services', 'autostart' ]
```
all is well, except for the top-level package would also require `-services`. I then ban it using `requires_ban`, it no longer does, but `-services` package is still brought in by `-bin` because [this](https://github.com/clearlinux/autospec/blob/master/autospec/specfiles.py#L232).

The workaround is to move all the systemd files into `server_extras` (which is what I've done). It results in more files named explicitly in autospec config, empty `-services` subpackage, unnecessary dependencies on that empty subpackage etc.

This approach (cutting "automatic" dependencies chains instead) seems a more reasonable to me. Especially since there's no existing case that breaks. It leverages existing subpackages without moving the files around and simpler config (i.e. the `server_extras` above will work as intended whereas right now it does not).

bryteise:
Okay, to me it seemed like just putting all the systemd files into server_extras was the correct thing to do for this case because the package is effectively two different applications (a client and a server) in the same source package and we don't want to always add the server if folks just want the client. I don't really see any issue with the empty services subpackage or the extra dependencies that don't have content.

To me making requires_ban (which really is only supposed to interact at the top level) go into the subpackages is odd. I think I could see somebody wanting requires_ban to be specific per subpackage but that gets really messy to me too.

I do like the smaller config but I don't really think for this specific case it is correct because those components really belong to the server side not the main package.

busykai:
OK. I'll close this PR.

PR #369, opened 2019-04-17
Set autoreconf when patch touches aclocal.m4

sunnyflunk:
While less common, changes to aclocal.m4 also require configure to be
regenerated.


bryteise:
Ah indeed, thanks =).

PR #370, opened 2019-04-18
license: Ignore file if couldn't be decoded

lpereira:
libcdio-2.1.0 has test files named "copying.iso" and "copying-rr.iso",
which are of course not text files -- but autospec considers those
to be license files.

### DIFF
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -76,6 +76,23 @@ def add_license(lic):
     return result
 
 
+def decode_license(license):
+    def try_with_charset(license, charset):
+        if not charset:
+            return
+
+        try:
+            return license.decode(charset)
+        except UnicodeDecodeError:
+            if charset in ('ISO-8859-1', 'ISO-8859-15'):
+                if b'\xff' in license:
+                    return try_with_charset(license, 'ISO-8859-13')
+                if b'\xd2' in license and b'\xd3' in license:
+                    return try_with_charset(license, 'mac_roman')
+
+    return try_with_charset(license, chardet.detect(license)['encoding'])
+
+
 def license_from_copying_hash(copying, srcdir):
     """Add licenses based on the hash of the copying file."""
     data = tarball.get_contents(copying)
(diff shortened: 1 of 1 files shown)


autospec/license.py
```diff
@@ -99,7 +99,11 @@ def license_from_copying_hash(copying, srcdir):
         # This is not a text file
         return
 
-    data = data.decode(license_charset)
+    try:
+        data = data.decode(license_charset)
```
phmccarty: Hmm, if `license_charset` is set to `ISO-8859-13` or `mac_roman` (see code a few lines above), I wouldn't expect a `UnicodeDecodeError` for those cases...

lpereira: If the code above these lines was working, then this will continue to work.  However, the automatic charset detection failed in this case (it's an ISO file, and charset thinks it's Windows-1251 IIRC), causing `data.decode()` to fail.

phmccarty: Yes, I see in this case `charset.detect` detected `Windows-1254`, but with a relatively low confidence score -- approximately `0.46` for `test/data/copying-rr.iso`. `1.0` is the highest score. Print out the value of `detected` to see the details.

We do not inspect the confidence score, but perhaps we should, and filter out files where the confidence is below a certain threshold?

phmccarty: After collecting more data, I see that some legitimate license files can have a low confidence score (e.g. `libs/icu/icu-src/LICENSE` from the texlive package is `ISO-8859-1` with a confidence of `0.73`), but other non-license files may have similar score (e.g. `test/data/copying.iso` from libcdio is `ISO-8859-1` with a confidence of `0.7291`). So, I don't trust the chardet confidence ratings enough for autospec to rely on them.

lpereira: Thanks for looking into this.  I suspected that this would be the case, given the heuristics  to change the detected charset to `ISO-8859-13` and `mac_roman` above.  Maybe those could be moved to the `except` block below, and only print the warning/return if those are not contemplated by the `if license_charset ...` block?

phmccarty: @lpereira Yes, I like that idea.

lpereira:
I've pushed a new version that does this. It also moves the code to
calculate the hash only after it successfully decodes the license text, so
it should be slightly faster on packages with large files.

autospec/license.py
```diff
@@ -76,31 +76,38 @@ def add_license(lic):
...
+            return
+
+        try:
+            return data.decode(license, charset)
+        except UnicodeDecodeError:
+            if license_charset in ('ISO-8859-1', 'ISO-8859-15'):
+                if b'\xff' in data:
+                    return try_with_charset(data, 'ISO-8859-13')
+                if b'\xd2' in data and b'\xd3' in data:
+                    return try_with_charset(data, 'mac_roman')
```
phmccarty: `data` is not defined in this function... I presume the correct variable name is `license`?

lpereira: Oops.

phmccarty:
@lpereira Sorry for the delayed review. Travis is still failing:

```
$ make check
autospec/license.py:87:16: F821 undefined name 'license_charset'
```

After this is fixed, and the PR rebased to latest master, I think it will be ready to merge.

lpereira:
Fixed and force-pushed.  `make check` is passing locally.  (Sorry about this -- I've written the first version on my Clear system, but the SSH keys to push to Github are in another computer.  No excuse for sending a PR without testing, though.)

lpereira:
Finally unit tests are passing. :)

bryteise:
Looks good to me, thanks!

PR #371, opened 2019-04-25
Ignore for-review.txt

phmccarty:
The Clear Linux OS developer tooling recently gained the ability to
create trimmed-down patches from autospec commits, dumping them to files
named "for-review.txt".

Update .gitignore to ignore these files.


phmccarty:
No, because autospec overwrites the file.

PR #372, opened 2019-04-26
Do not print the same license twice

lpereira:
For large pieces of software that brings in a lot of dependencies, it
takes quite a bit of time to fetch the license information from the
license server.

Many of the "unknown licenses" messages end up buried as a result.

This patch caches the hash of a license in a set maintained by
license_from_copying_hash().

### DIFF
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -76,7 +76,7 @@ def add_license(lic):
     return result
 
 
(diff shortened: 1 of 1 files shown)


lpereira:
Ah, I see.  Closing the PR, then.

PR #373, opened 2019-04-30
Set ${GCC_IGNORE_WERROR} environment variable

lpereira:
Clear ships with a patch in GCC that allows ignoring the -Werror
compilation flag if this environment variable is set.  -Werror is a
useful flag for the upstream package maintainers, but is a source of
headaches for downstream users.


lpereira:
Rebased and force-pushed.

PR #374, opened 2019-05-03
Add the ability to not auto-unpack archives

thiagomacieira:
Usually needed when it's a .zip file without a prefix, which results
in weird mv errors when launching Autospec. This also makes it
possible to have more sources (like .jar).


chivakker:
lgtm, tested locally with R-rstudio package.

PR #380, opened 2019-05-20
Updates

fenrus75:
various updates that have been well tested locally

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -182,7 +182,7 @@ class Specfile(object):
         for pkg in sorted(self.packages):
             if pkg == "autostart" and self.no_autostart:
                 continue
-            if pkg.endswith("-extras"):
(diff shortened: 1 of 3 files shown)


autospec/specfiles.py
```diff
@@ -1326,6 +1322,8 @@ def write_qmake_pattern(self):
             extra_qmake_args = "-spec linux-clang "
         if config.config_opts['use_lto']:
             extra_qmake_args += "-config ltcg "
+        else:
+            flags.extend(["-fno-lto"])
```
bryteise: Should this be ```extras_qmake_args += "-fno-lto " instead? (flags isn't a think in the write_qmake_pattern). Or does this need to exist at all since the arg isn't being passed? Maybe @thiagomacieira can give me a little insight here.

bryteise:
Using #384 To update this PR.

PR #387, opened 2019-05-28
Lto clang fixes

bryteise:
(also throw in a readme fix)

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -448,6 +448,10 @@ class Specfile(object):
             self._write_strip("export CC=clang\n")
             self._write_strip("export CXX=clang++\n")
             self._write_strip("export LD=ld.gold\n")
+            lto = "-flto"
+        else:
+            lto = "-flto=4"
+
         if config.config_opts['optimize_size']:
             if config.config_opts['use_clang']:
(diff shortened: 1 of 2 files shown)


autospec/specfiles.py
```diff
@@ -482,18 +486,23 @@ def write_variables(self):
                 flags.extend(["-O3", "-fno-semantic-interposition", "-falign-functions=32", "-fno-math-errno", "-fno-trapping-math"])
         if self.default_pattern != 'qmake':
             if config.config_opts['use_lto']:
-                flags.extend(["-O3", "-flto=4", "-ffat-lto-objects"])
-                self._write_strip("export AR=gcc-ar\n")
-                self._write_strip("export RANLIB=gcc-ranlib\n")
-                self._write_strip("export NM=gcc-nm\n")
+                flags.extend(["-O3", lto, "-ffat-lto-objects"])
+                if not config.config_opts['use_clang']:
```
phmccarty: Shouldn't this line check the opposite case, that is `if config.config_opts['use_clang']` ?

bryteise: Ugh yes, thanks for catching that one. I changed what I was doing there a few times and lost track it seems of the logic X(.

PR #390, opened 2019-05-29
Ban any dependency on 'futures'

phmccarty:
'futures' provides concurrent.futures, which has been part of the Python
standard library since the 3.2 release, so it should not be added as a
dependency anymore.

Any remaining Python 2 packages can manually require
'deprecated-futures-legacypython' as needed.


bryteise:
Thanks!

PR #394, opened 2019-05-31
Fix zip archive scanning

### LINKED ISSUE #393: build_unzip scans wrong line of zip info for prefix
build_unzip() calls unzip -q -l to get a listing of zip file contents. It then expects to find the first file listed at the 5th line (4th if no hash). But -q suppresses the line listing the archive name, so it should be looking at the 4th (3rd if no hash) line instead.

Patch forthcoming.

bwarden:
ZIP info with -q starts on the 3rd line, not the 4th. In the case of an
archive with a single file, this will fail.

Fixes #393


bryteise:
Ah thanks for the fixup and test update!

PR #395, opened 2019-05-31
Update format of custom_extras files

bryteise:
Switching to flat file handling using the same conf file parsing as
other config files like extras and dev_extras.

Moves to two different config files for the file list and the requires
list.

### DIFF
--- a/autospec/git.py
+++ b/autospec/git.py
@@ -60,7 +60,6 @@ def commit_to_git(path):
     call("bash -c 'shopt -s failglob; git add -f *.sha256'", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("bash -c 'shopt -s failglob; git add -f *.sign'", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("bash -c 'shopt -s failglob; git add -f *.pkey'", check=False, stderr=subprocess.DEVNULL, cwd=path)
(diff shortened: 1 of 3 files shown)


autospec/git.py
```diff
@@ -60,7 +60,6 @@ def commit_to_git(path):
     call("bash -c 'shopt -s failglob; git add -f *.sha256'", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("bash -c 'shopt -s failglob; git add -f *.sign'", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("bash -c 'shopt -s failglob; git add -f *.pkey'", check=False, stderr=subprocess.DEVNULL, cwd=path)
-    call("bash -c 'shopt -s failglob; git add -f *_extras'", check=False, stderr=subprocess.DEVNULL, cwd=path)
```
phmccarty: Don't we still want to commit any files that match `*_extras`? Though it needs an extension to `*_extras*` now.

bryteise: This is something that is handled via read_conf_file, when foo_extras and foo_extras_requires is read they are added to the set of config_files that git manages.

phmccarty: Ah, yes, I see that now, thanks.

bryteise:
@thiagomacieira This should make life easier for updating the extras files.

PR #397, opened 2019-06-05
meson: add a missing space between buildtype and configure extras

chivakker:
fixing a typo when building for avx2, moved first argument to upper
line to make it more evident

Signed-off-by: Daniel Charles <daniel.charles@intel.com>

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -1495,8 +1495,8 @@ class Specfile(object):
 
         if config.config_opts['use_avx2']:
             self._write_strip('CFLAGS="$CFLAGS -m64 -march=haswell" CXXFLAGS="$CXXFLAGS -m64 -march=haswell " LDFLAGS="$LDFLAGS -m64 -march=haswell" '
-                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain'
-                              '{0} {1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
+                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain {0} '
+                              '{1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
             self._write_strip('ninja -v -C builddiravx2')
 
         if config.config_opts['32bit']:


chivakker:
@bryteise this is my fix for the configure flags problem.

bryteise:
Ahhh that indeed makes it obvious, thanks for that fix!

autospec/specfiles.py
```diff
@@ -1495,8 +1495,8 @@ def write_meson_pattern(self):
 
         if config.config_opts['use_avx2']:
             self._write_strip('CFLAGS="$CFLAGS -m64 -march=haswell" CXXFLAGS="$CXXFLAGS -m64 -march=haswell " LDFLAGS="$LDFLAGS -m64 -march=haswell" '
-                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain'
-                              '{0} {1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
+                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain ${0) '
```
phmccarty: On second glance, I'm not understanding the addition to this line. Should it be ` {0} ` rather than ` ${0) `?

bryteise: Ah indeed, good catch. Looks like shell syntax snuck into the fingers.

chivakker: updated this PR with the catch fixed.

autospec/specfiles.py
```diff
@@ -1495,8 +1495,8 @@ def write_meson_pattern(self):
 
         if config.config_opts['use_avx2']:
             self._write_strip('CFLAGS="$CFLAGS -m64 -march=haswell" CXXFLAGS="$CXXFLAGS -m64 -march=haswell " LDFLAGS="$LDFLAGS -m64 -march=haswell" '
-                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain'
-                              '{0} {1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
+                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain {0) '
```
phmccarty: There is still a syntax issue: replace the right parenthesis with a right curly bracket (i.e. should be `{0}`)

chivakker: oh boy, tried again, thanks for catching.

PR #398, opened 2019-06-07
Add translation for importlib_metadata

phmccarty:
The module name is 'importlib_metadata', but the package name is 'importlib-metadata', so this needs a translation.


bryteise:
Hrm I would like to look closer as to why 311d4f4792420aa901b9a8d87cef4b9a0f3a612c isn't making this translation unnecessary.

bryteise:
Actually it looks like the package name should have the underscore so I'm doubly confused.

phmccarty:
Disregard this request.

PR #399, opened 2019-06-07
Add mvnbin build pattern

AthenasJimenez:
This pattern would add the ability to create bootstraps specfiles for
maven dependencies(aka artifacts). These boostraps would be use as a
way to solve the dependencies for Java projects that use Maven as a
build tool.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -589,6 +589,27 @@ class Specfile(object):
 
         self.write_find_lang()
 
+    def write_mvnbin_install(self):
+        patterns = [
+            re.compile(r"^https://repo1.maven.org/maven2/([a-z\-]+)/([a-z\-]+)/([\d.]+)/[a-z-.\d]*\.[pom|jar]")]
+        mvn_sources = [self.url] + self.sources["archive"]
+        src_num = 0
+
+        for src in mvn_sources:
+            for pat in patterns:
+                m = pat.match(src)
+                if m:
+                    groupid = m.group(1)
+                    artifactid = m.group(2)
+                    version = m.group(3)
+                    src_dir = "{}/{}/{}".format(groupid, artifactid, version)
+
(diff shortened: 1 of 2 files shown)


autospec/tarball.py
```diff
@@ -501,6 +501,8 @@ def find_extract(tar_path, tarfile):
         extract_cmd, tar_prefix = build_unzip(tar_path)
     elif tarfile.lower().endswith('.gem'):
         extract_cmd, tar_prefix = build_gem_unpack(tar_path)
+    elif tarfile.lower().endswith('.jar'):
```
bryteise: Is this handling alternate to maven java build systems case?

autospec/specfiles.py
```diff
@@ -1495,8 +1528,8 @@ def write_meson_pattern(self):
 
         if config.config_opts['use_avx2']:
             self._write_strip('CFLAGS="$CFLAGS -m64 -march=haswell" CXXFLAGS="$CXXFLAGS -m64 -march=haswell " LDFLAGS="$LDFLAGS -m64 -march=haswell" '
-                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain {0} '
-                              '{1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
+                              'meson --prefix /usr --libdir=/usr/lib64/haswell --buildtype=plain'
+                              '{0} {1} builddiravx2'.format(config.extra_configure, config.extra_configure64))
```
bryteise: This shouldn't be in this change, please rebase.

bryteise:
Looks pretty good to me other than the extra bit that needs to be rebased out.

autospec/tarball.py
AthenasJimenez: No, this is needed by the maven bin case to get the License info and description.

autospec/specfiles.py
AthenasJimenez: O gosh, forgot to check for new changes. I'll fix it.

bryteise:
Great, thanks!

PR #402, opened 2019-06-11
Improve go ecosystem support

bryteise:
Update support for handling go packages by handling missing go
dependencies. Make use of go modules and the go proxy (see
https://tip.golang.org/cmd/go/#hdr-Module_proxy_protocol for more
details on the proxy implementation) to discover what artifacts are
needed.

A new go-dep build_pattern that will package all versions that the
go-proxy finds is also added as part of this change. Additional rework
of the golang build_pattern (and Makefile build_pattern as a number of
go packages use Makefiles) to use module vendoring when module
configuration is detected was added as well.

The new functionality is also intended to be used by a higher level
tool that will wrap autospec calls to build missing dependencies as
well as the requested package.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -184,6 +184,50 @@ def is_version(num_str):
     return False
 
 
+def parse_go_mod(path):
+    """Parse go.mod file for build requirements.
+
+    File content looks as follows:
+
+    module example.com/foo/bar
+
+    require (
+        github.com/BurntSushi/toml v0.3.1
+        git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999
+        github.com/inconshreveable/mousetrap v1.0.0 // indirect
+        "github.com/spf13/cobra" v0.0.3
+        github.com/spf13/pflag v1.0.3 // indirect
+    )
+
+    Need to handle all require lines including //indirect.
+    Skip requires that use .git for now. May need to be handled
+    differently.
+    """
+    reqs = []
+    with open(path, "r") as gfile:
+        dep_start = False
+        for line in gfile.readlines():
+            # Ideally the mod file is generated and the format is
+            # always correct but add a few defenses just in case
+            line = line.strip()
+            if line.startswith("//"):
+                # Skip comments
+                continue
+            if dep_start:
+                # End of the require section
+                if line.startswith(")"):
+                    break
+                req = line.split()[:2]
+                req[0] = req[0].replace('"', '')
+                if req[0].endswith(".git"):
+                    continue
+                reqs.append(req)
+                continue
+            if line.startswith("require ("):
+                dep_start = True
+    return reqs
+
+
 def parse_modules_list(modules_string, is_cmake=False):
     """Parse the modules_string for the list of modules, stripping out the version requirements."""
     if is_cmake:
@@ -262,7 +306,8 @@ def parse_cargo_toml(filename):
 
 
 def _get_desc_field(field, desc):
-    """
+    """Get a given field in an R DESCRIPTION file.
+
     Internal helper to read the value for the given field in an R package
     DESCRIPTION file. Returns a list representing the value; if value contained
     commas, the string is split on the commas and surrounding whitespace to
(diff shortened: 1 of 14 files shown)


bryteise:
@phmccarty Updated though I should note this is not fully functional for go packages that have dependencies not using go modules.

autospec/tarball.py
```diff
@@ -43,6 +45,32 @@
...
+        if not os.path.exists(path):
+            download.do_curl(os.path.join(url, name),
+                             dest=path,
+                             is_fatal=True)
+        sha1 = get_sha1sum(path)
+        write_upstream(sha1, name, mode="a")
+
+
+def process_go_dependency(url, target):
+    """Handle go dependiency files."""
```
phmccarty: s/dependiency/dependency

autospec/tarball.py
```diff
@@ -573,18 +651,22 @@ def process(url_arg, name_arg, ver_arg, target, archives_arg, filemanager):
...
-    name_and_version(name_arg, ver_arg, filemanager)
     # determine build pattern and build requirements from url
     detect_build_from_url(url)
+    # Create the download path for content and set build.download_path
+    create_download_path(target)
+    # determine name and version of package
+    name, rawname, version = name_and_version(name_arg, ver_arg, filemanager)
+    # '-' aren't allowed in the version string for spec files
+    # rpmbuild will fail so replace them with '.'
+    if '-' in version:
```
phmccarty: Can you move this version mangling code into a separate function? Maybe into util.py?

autospec/tarball.py
```diff
@@ -573,18 +651,22 @@ def process(url_arg, name_arg, ver_arg, target, archives_arg, filemanager):
...
+        version = version.replace('-', '.')
     # set gcov file information, must be done after name is set since the gcov
     # name is created by adding ".gcov" to the package name (if a gcov file
     # exists)
     set_gcov()
     # download the tarball to tar_path
-    tar_path = download_tarball(target)
-    # write the sha of the upstream tarfile to the "upstream" file
-    write_upstream(get_sha1sum(tar_path), tarfile)
+    tar_path = check_or_get_file(url, os.path.basename(url))
```
phmccarty: I think the second argument here can be `tarfile`, correct?

autospec/git.py
```diff
@@ -117,6 +119,7 @@ def commit_to_git(path):
         "!*.tar.*.*",
         "*.zip",
         "commitmsg",
+        "list",
```
phmccarty: The filename `list` is not very descriptive... Can it have a different name that better reflects its purpose?

autospec/buildreq.py
```diff
@@ -738,8 +783,23 @@ def scan_for_configure(dirn):
         if any(f.endswith(".go") for f in files):
             add_buildreq("buildreq-golang")
             buildpattern.set_build_pattern("golang", default_score)
-            if "go.mod" in files:
-                config.set_gopath = False
+
+        if "go.mod" in files:
```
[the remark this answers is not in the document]
bryteise: Goodness unfortunately this is pushing my memory but I do remember originally I had that condition set and I think I removed it due to this not being the case. I am not opposed to adding it back though as the sets of things supported by this change has decreased since then.

autospec/buildreq.py
```diff
@@ -184,6 +184,50 @@ def is_version(num_str):
...
+        for line in gfile.readlines():
+            # Ideally the mod file is generated and the format is
+            # always correct but add a few defenses just in case
+            line = line.strip()
+            if line.startswith("//"):
+                # Skip comments
+                continue
+            if dep_start:
+                # End of the require section
+                if line == ")":
```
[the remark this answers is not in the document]
bryteise: I don't know if it supported or not but the go.mod file is intended to be managed by the go mod command and it won't do either this or the following comment on requires paren lines. I think technically the syntax supports ```require ( thing )``` but again this isn't how go mod will organize the file. I don't really see a downside to checking against startswith though so I'll change it.

autospec/tarball.py
bryteise: Fixing.

autospec/tarball.py
bryteise: Indeed, fixing.

autospec/git.py
bryteise: Actually because of how I am handling the list file generation I can likely rework even downloading the file. Thanks for pointing this out. Fixing.

autospec/tarball.py
```diff
@@ -248,12 +298,10 @@ def download_tarball(target_dir):
...
     build.download_path = target
     call("mkdir -p {}".format(build.download_path))
-
-    # locate the tarball locally or download
-    return check_or_get_file(url, tarfile)
+    return target
 
 
-def convert_version(ver_str):
+def convert_version(ver_str, name):
```
phmccarty: The new `name` parameter here is unused in the function. Is it supposed to be used, or is it left over from a previous iteration of the PR?

bryteise: It is being used, see line 323. This is a weird change though because I'm masking a global so you don't see any source change in the function. Would you prefer I used a different variable name and update the source to use that?

phmccarty: No, sorry, I missed the variable references within the function. My comment was invalid.

PR #406, opened 2019-06-20
build: skip blank lines when populating file list

phmccarty:
If blank lines appear in the file list from the build log, skip them.


phmccarty:
For the record, I should mention that this fix also resolves a persistent autospec warning I am seeing when testing the latest release of `mock`:

```
[WARNING] Removing directory  from file list
```

The empty string in the above warning message originates from the empty string being added to the files list, as two blank lines appear after the files list from the build log.

PR #410, opened 2019-06-21
specfiles.py: export LANG=C.UTF-8

jurobystricky:
Generate spec files with:

export LANG=C.UTF-8

instead of the current

export LANG=C

to avoid occasional build warnings such as:

msggrep: warning: Locale charset "ANSI_X3.4-1968" is different from
                  input file charset "UTF-8".
                  Output of 'msggrep' might be incorrect.
                  ...


bryteise:
@phmccarty This seems safe to me, any objections?

PR #412, opened 2019-06-24
Add open_auto helper function

phmccarty:
This helper function is meant to be used throughout autospec as a replacement for most uses of open(), especially whenever reading arbitrary data from upstream sources.

It reads/writes data in UTF-8 always, with invalid UTF-8 characters escaped with Python's "surrogate" escaping mechanism. This ensures that any Latin-1 or other non-UTF-8 compatible encoded data can successfully be read in and later written out without data corruption.

### DIFF
--- a/autospec/abireport.py
+++ b/autospec/abireport.py
@@ -197,7 +197,7 @@ def truncate_file(path):
     """Zero file content."""
     if not os.path.exists(path):
         return
-    with open(path, "r+", encoding="utf-8") as trunc:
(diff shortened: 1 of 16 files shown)


bryteise:
Looks great, can I go ahead and merge it now?

phmccarty:
Yes, I have tested a few cases where we currently see corruption, and the issues disappear with the changes from this PR.

bryteise:
Thanks!

Issue #134, opened 2018-03-14
'packages' file fails to be added to package git repo


phmccarty:
Whenever I run autospec with the `git` option enabled in autospec.conf, it prints this error message right before the git commit output:

```
fatal: pathspec 'packages' did not match any files
```

Briefly looking at the code (autospec/git.py), I think it's trying to `git add` the file pointed to by the autospec.conf `packages_file` option. But the packages file I've configured autospec to use does not live in the repo, so the command fails.

phmccarty:
Fixed by #413

PR #413, opened 2019-06-24
config: skip tracking of the 'packages' file

phmccarty:
This 'packages' file looks like a config file, but because it typically exists outside the package git repo, we want avoid a later attempt to commit it.

This change also makes the "fatal: pathspec 'packages' did not match any files" error message disappear.


phmccarty:
Fixes #134

bryteise:
Lovely, thanks!

PR #415, opened 2019-06-26
Some pkg_integrity cleanups

phmccarty:
* Support current default and legacy/mirror pypi URLs
* Remove get_hash_url() to ease the process of adding additional verifiers
* Clean up progress a bit


bryteise:
Looks reasonable to me. Do you expect to add the qt verification class to this PR or is it complete as is?

phmccarty:
I think I want to enable the Qt verification first, since that has been a long outstanding gap.

phmccarty:
@bryteise I implemented the Qt sha256 verifier.

Part way through, I remembered the previous PR to resolve this (#319) but opted to just add a separate verifier class. The new class is relatively small, though it would be nice eventually to refactor for even more code sharing. (At least the sha verification code is already split out into ShaSumVerifier, so that helps.)

bryteise:
Looks good, thanks!

PR #417, opened 2019-07-01
Implement new config for clr-service-restart units

phmccarty:
To improve tracking of units that are enabled for restart by clr-service-restart, add a new config file `service_restart`. It lists one or more unit names (with full path), one unit per line. The %install section is updated accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -347,6 +347,12 @@ golang_libpath
   This could be easily fixed by placing ``gopkg.in/yaml.v`` in this file,
   changing where the go bits will be placed.
 
+service_restart
+  Each line in the file specifies the full path to a systemd unit file
+  installed by this package that should be restarted by clr-service-restart_.
+
+.. _clr-service-restart: https://github.com/clearlinux/clr-service-restart
+
 Controlling files and subpackages
 ---------------------------------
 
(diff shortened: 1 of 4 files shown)


phmccarty:
Whoops, the symlink destinations are wrong. Will update shortly.

fenrus75:
If the common case is one service... should there be an options.conf
Boolean for this?

phmccarty:
I considered that approach, but the implementation becomes more complex because it needs to tie into the build rounds, parsing the build log, etc.

autospec/git.py
```diff
@@ -54,6 +54,7 @@ def commit_to_git(path):
     call("git add make_prepend", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("git add install_prepend", check=False, stderr=subprocess.DEVNULL, cwd=path)
     call("git add install_append", check=False, stderr=subprocess.DEVNULL, cwd=path)
+    call("git add service_restart", check=False, stderr=subprocess.DEVNULL, cwd=path)
```
bryteise: This isn't required because read_conf_file adds it to the config_files list and that entire list is added above.

bryteise:
Change looks good aside from my one comment.

autospec/git.py
phmccarty: Ah, thanks for the catch. I have fixed this.

bryteise:
Thanks!

PR #418, opened 2019-07-02
Parse CRAN package descriptions for dependencies

phmccarty:
A DESCRIPTION file is present in every CRAN package, and dependency information is listed in the Imports value, and (for legacy packages) the Depends value.

Some of the listed packages are part of R itself, but others are simply missing from distro, so also print a "skipping" message for those cases.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -261,6 +261,44 @@ def parse_cargo_toml(filename):
             add_requires(cdep)
 
 
+def _get_desc_field(field, desc):
+    """
+    Internal helper to read the value for the given field in an R package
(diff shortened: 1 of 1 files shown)


autospec/buildreq.py
```diff
@@ -261,6 +261,44 @@ def parse_cargo_toml(filename):
             add_requires(cdep)
 
 
+def _get_desc_field(field, desc):
```
bryteise: Oh look at us starting to try and indicate private functions, I should start adopting this pattern X(.

bryteise:
Nice.

PR #421, opened 2019-07-02
Support UTF-8 curly single quotes for R patterns

phmccarty:
Since the introduction of the C.UTF-8 locale for builds, R now prints
left/right curly single quotes in its error messages.

Adapt to this behavior by accepting U+2018 (left single quotation mark)
and U+2019 (right single quotation mark) in addition to the apostrophes
for the R fail patterns.


bryteise:
Hah, that's a fun one. Thanks!

PR #423, opened 2019-07-03
maven pattern: Mapping more maven packages

AthenasJimenez:


bryteise:
This looks like something that is maybe a temporary work around while we are getting the ecosystem in place?

AthenasJimenez:
Yes, eventually I'll  replace this document with sub packages and/or Provides tags on specfiles.

PR #425, opened 2019-07-03
buildreq: ignore R packages that ship with R itself

### LINKED ISSUE #420: buildreq: ignore R packages that are part of R itself
With the changes from #418, dependencies of CRAN packages are now better detected. To build on this approach, autospec should also track which packages are part of `R` itself and avoid printing "skipped" progress messages for those.

There is likely a way to autodetect this list of packages, but we can start with a package-name blacklist and work from there.

phmccarty:
Add a blacklist of packages that ship with R so that autospec can ignore them and avoid printing a "skipping" progress message for them.

Eventually, we want to autopopulate this list, so I added a FIXME too.

Fixes #420

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -281,6 +281,44 @@ def _get_desc_field(field, desc):
     return val
 
 
+# FIXME: this list should be autodetected
+def _get_r_provides():
+    """Set of packages provided by R, serving as blacklist for CRAN deps."""
+    provides = [
+        "KernSmooth",
+        "MASS",
+        "Matrix",
+        "base",
+        "boot",
+        "class",
+        "cluster",
+        "codetools",
+        "compiler",
+        "datasets",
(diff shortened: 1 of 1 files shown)


autospec/buildreq.py
```diff
@@ -281,6 +281,44 @@ def _get_desc_field(field, desc):
...
+        "stats4",
+        "survival",
+        "tcltk",
+        "tools",
+        "translations",
+        "utils",
+    ]
+    return pkg in provides
+
+
```
bryteise: Since this is in a loop I wonder if a _get_r_provides function that returns a **set** (and is only called once) that is then used in the loop would be a better way of doing this.

phmccarty: Yes, good idea... v2 patch will appear shortly.

bryteise:
Looks good, thanks!

PR #428, opened 2019-07-11
download: set pycurl timeouts

phmccarty:
To start with, try setting the following timeouts:

- connection timeout of 10 seconds
- request timeout of 10 seconds
- speed limit timeout of 1 byte/sec in a 10-second window

Feel free to suggest different defaults.


bryteise:
Looks good, just need some mock fixes for the tests.

PR #430, opened 2019-07-12
download: Increase pycurl timeouts to 30 seconds

### LINKED ISSUE #429: pycurl: timeout of 10 seconds is quite demanding
New timeouts introduced by commit 8d99b0f8d8fe4264d3e9ec756c13fcb1ad86ef82 is causing timeouts for our use case.

Can it be increased to more than 30 seconds?

mandarcthorat1:
Increased connection timeout, request timeout and speed limit timeout
window to 30 seconds from 10 seconds

Fixes #429 

Signed-off-by: Mandar Thorat <mandar.chandrakant.thorat@intel.com>

### DIFF
--- a/autospec/download.py
+++ b/autospec/download.py
@@ -44,10 +44,10 @@ def do_curl(url, dest=None, post=None, is_fatal=False):
         c.setopt(c.POSTFIELDS, post)
     c.setopt(c.FOLLOWLOCATION, True)
     c.setopt(c.FAILONERROR, True)
-    c.setopt(c.CONNECTTIMEOUT, 10)
-    c.setopt(c.TIMEOUT, 10)
+    c.setopt(c.CONNECTTIMEOUT, 30)
+    c.setopt(c.TIMEOUT, 30)
     c.setopt(c.LOW_SPEED_LIMIT, 1)
-    c.setopt(c.LOW_SPEED_TIME, 10)
+    c.setopt(c.LOW_SPEED_TIME, 30)
     buf = BytesIO()
     c.setopt(c.WRITEDATA, buf)
     try:
--- a/autospec/download.py
+++ b/autospec/download.py
@@ -44,10 +44,10 @@ def do_curl(url, dest=None, post=None, is_fatal=False):
         c.setopt(c.POSTFIELDS, post)
     c.setopt(c.FOLLOWLOCATION, True)
     c.setopt(c.FAILONERROR, True)
-    c.setopt(c.CONNECTTIMEOUT, 30)
+    c.setopt(c.CONNECTTIMEOUT, 10)
     c.setopt(c.TIMEOUT, 30)
     c.setopt(c.LOW_SPEED_LIMIT, 1)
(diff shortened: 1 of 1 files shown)


autospec/download.py
```diff
@@ -44,10 +44,10 @@ def do_curl(url, dest=None, post=None, is_fatal=False):
         c.setopt(c.POSTFIELDS, post)
     c.setopt(c.FOLLOWLOCATION, True)
     c.setopt(c.FAILONERROR, True)
-    c.setopt(c.CONNECTTIMEOUT, 10)
-    c.setopt(c.TIMEOUT, 10)
+    c.setopt(c.CONNECTTIMEOUT, 30)
```
bryteise: It seems a little much to me to increase this timeout to 30s. From the [docs](https://curl.haxx.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html) this is the timeout for how long the initial connect takes. Are you experiencing connections that take longer than 10s to establish?

autospec/download.py
```diff
@@ -44,10 +44,10 @@ def do_curl(url, dest=None, post=None, is_fatal=False):
...
+    c.setopt(c.TIMEOUT, 30)
     c.setopt(c.LOW_SPEED_LIMIT, 1)
-    c.setopt(c.LOW_SPEED_TIME, 10)
+    c.setopt(c.LOW_SPEED_TIME, 30)
```
bryteise: Similarly, 30s of being 1 byte per second or lower seems a bit much to me too. Given that you moved the absolute timeout to 30s, I don't think this would have an impact either way though.

autospec/download.py
mandarcthorat1: @bryteise The current value of 10s for CONNECTTIMEOUT works for my use case. I have proposed a  higher value of 30s for this parameter as its default timeout value is about 300s. 
Will it be fine to have a relaxed timeout for retries if curl request fails due to timeout error.

phmccarty:
@mandarcthorat1 Which of the three timeouts are you encountering? I am not surprised that the timeouts are a bit aggressive, but I would rather identify which timeouts are too short and work from there, rather than increasing them all by the same amount at the same time.

autospec/download.py
mandarcthorat1: @bryteise I completely agree with your reasoning here.

mandarcthorat1:
Hi @phmccarty,

I had encountered [CURLOPT_TIMEOUT](https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html) multiple times today which is currently set to 10s.

phmccarty:
@mandarcthorat1 Okay, because the increased `CURLOPT_TIMEOUT` is helping the most for you, can we compromise by only changing that value from `10` to `30`?

mandarcthorat1:
Hi @phmccarty,

It's a good compromise to make as currently nobody is reporting issues with other timeout values. I have pushed a commit that reverts changes to timeout values other than `CURLOPT_TIMEOUT`.

PR #436, opened 2019-07-24
cmake_modules: Set openjdk11 as default.

AthenasJimenez:
Openjdk9 is no longer under development and would be deprecated on July
26 of 2019 on Clear Linux. So, it would be replaced by Openjdk11
which is the current LTS.


bryteise:
Thanks!

PR #439, opened 2019-07-31
mvnbin pattern: Manage metadata

AthenasJimenez:
Add the capability to manage maven-metadata.xml files.


AthenasJimenez:
@phmccarty would you mind to take a look into this? Since William is not around :(

Issue #239, opened 2018-08-29
Add ability to add custom Summary field


phmccarty:
The spec file Summary field is always autogenerated at the moment. It would be nice to support adding a custom Summary, similar to how a custom %description is supported (for the base package) via the `description` file.

phmccarty:
@ahkok has a patch to implement: https://lists.clearlinux.org/pipermail/dev/2019-August/003351.html

phmccarty:
Resolved with #442

PR #442, opened 2019-08-02
Add `summary` file support, just like `description`.

ahkok:
/trivial/

### DIFF
--- a/autospec/specdescription.py
+++ b/autospec/specdescription.py
@@ -274,8 +274,11 @@ def scan_for_description(package, dirn):
 
 def load_specfile(specfile):
     """Load specfile with parse results."""
-    specfile.default_sum = default_summary
     if config.custom_desc:
         specfile.default_desc = "\n".join(config.custom_desc)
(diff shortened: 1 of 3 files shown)


ahkok:
Fixes #239

autospec/specdescription.py
```diff
@@ -274,8 +274,11 @@ def scan_for_description(package, dirn):
 
 def load_specfile(specfile):
     """Load specfile with parse results."""
-    specfile.default_sum = default_summary
     if config.custom_desc:
         specfile.default_desc = "\n".join(config.custom_desc)
     else:
         specfile.default_desc = default_description
+    if config.custom_summ:
+        specfile.default_sum = "\n".join(config.custom_summ)
```
phmccarty: Since `config.custom_summ` may contain multiple lines, we should just assign the first line and ignore any remaining lines.

ahkok: Fixed. Only first line is used (tested).

PR #445, opened 2019-08-13
Abirepot: omit -extras- subrpms.

ahkok:
We don't want to make bundles that include all the deps for *extras* rpms,
these are most likely removed or optional components. We may have to
create separate used_libs_extras files for them instead.

### DIFF
--- a/autospec/abireport.py
+++ b/autospec/abireport.py
@@ -27,6 +27,7 @@ import shutil
 import subprocess
 import sys
 
+import tarball
 import util
 
 valid_dirs = ["/usr/lib", "/usr/lib64"]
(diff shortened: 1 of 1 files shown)


autospec/abireport.py
```diff
@@ -219,8 +219,19 @@ def examine_abi(download_path):
 
 def examine_abi_host(download_path, results_dir):
     """Make use of the hostside abireport tool."""
+    rpms = set()
+    for item in os.listdir(results_dir):
+        if item.find("-extras-") > 0:
```
phmccarty: Checking for `-extras-` will also capture rpm names like `kio-extras-lib-19.04.3-29.x86_64.rpm`. To make it more robust, I suggest to account for the package name in the rpm name (see `tarball.name` global variable).

ahkok: works now.

PR #448, opened 2019-08-16
Create ant buildpattern

bwarden:
I think we're ready for review now. Any ant package will have to have the target specified in `make_args` and the installation completely written in `install_append`. There's just no dependable pattern beyond that.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -1558,7 +1558,6 @@ class Specfile(object):
         self._write_strip("ant -d -v " + self.extra_make)
         self._write_strip("%install")
         self.write_install_prepend()
-        jar_dir = os.path.join("%{buildroot}/usr/share/jar", self.name)
         self._write_strip("")
 
     def write_maven_pattern(self):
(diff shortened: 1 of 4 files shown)


bwarden:
Spoke too soon.

AthenasJimenez:
Looks good to me :)

autospec/specfiles.py
```diff
@@ -1555,12 +1555,9 @@ def write_ant_pattern(self):
         self.write_build_prepend()
         self.write_proxy_exports()
         self._write_strip("export ANT_HOME=/usr/share/ant")
-        self._write_strip("ant dist")
+        self._write_strip("ant -d -v " + self.extra_make)
         self._write_strip("%install")
         self.write_install_prepend()
-        jar_dir = os.path.join("%{buildroot}/usr/share/jar", self.name)
-        self._write_strip(f"mkdir -p {jar_dir}")
-        self._write_strip(f"install -m 0644 dist/*.jar {jar_dir}")
```
phmccarty: Are we going to use `install_append` exclusively for packages that build with Ant? Or is the install process TBD?

bwarden: Exclusively, yes. Artifacts produced vary too much, and for some packages, we'll want to install the JAR, while for others, we'll want to install regular files and _not_ the JAR.

phmccarty: /me reads the PR summary... sorry for reading that very carefully.

Sounds good to me.

PR #449, opened 2019-08-18
Add new pattern for 'vala'

jurobystricky:
Autospec fails for several gnome packages with:
meson.build:1:0: ERROR: Could not execute Vala compiler "valac"
This had to be fixed by manually adding "vala" to
buildreq_add file.
This patch adds "BuildRequires : vala" to the spec file if the
pattern "ERROR: Could not execute Vala compiler" is detected.


jurobystricky:
Pattern found in several gnome packages.
Patch tested  with:
<pre>
make autospecnew URL="https://codeload.github.com/GNOME/gnome-chess/tar.gz/3.33.90" NAME="gnome-chess"
</pre>

AthenasJimenez:
@mrkz any comments here?

PR #452, opened 2019-08-21
specfiles: add new function for installing license files

phmccarty:
Reduce code duplication by moving identical code blocks into a new function, `write_license_files()`.


AthenasJimenez:
+1

PR #457, opened 2019-08-30
Maven source buildpattern

bwarden:
Seems to be working at least for the champion package mvn-jline.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -638,26 +638,27 @@ class Specfile(object):
             src_num += 1
 
     def write_maven_install(self):
+        """Write installation steps to specfile for maven source packages."""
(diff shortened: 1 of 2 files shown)


bwarden:
I would prefer to squash before we merge, btw.

autospec/specfiles.py
```diff
@@ -637,6 +637,29 @@ def write_mvnbin_install(self):
                     break
             src_num += 1
 
+    def write_maven_install(self):
+        self.write_license_files()
+
+        """Figure out the artifact details and path components."""
```
phmccarty: More idiomatic is to use python comments beginning with `#` here, instead of docstrings. Same applies to the other comments you have added.

PR #458, opened 2019-09-01
license.py: Fix a misleading error message

jurobystricky:
We may get a message "<package>.license not found" even
if the file exists.

The <package>.license file is manually created but the
licenses in the file may be silently rejected due to being
blacklisted.
For example adding "GPL" or "BSD" to abc.license while autospecing
a package "abc" can produce the error:

"[FATAL]  Cannot find any license or abc.license file!"

eventhough the abc.license manifestly exists.

This patch modifies the original error message and also prints out
some additional messages, such as:

[WARNING] lapack.license: ambiguous license GPL ignored.
[WARNING] lapack.license: ambiguous license BSD ignored.
...
[FATAL]  Cannot find any license or a valid lapack.license file!

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -845,7 +845,8 @@ def parse_config_files(path, bump, filemanager, version):
         words = content[0].split()
         for word in words:
             if word.find(":") < 0:
-                license.add_license(word)
+                if not license.add_license(word):
+                    print_warning("{}: blacklisted license {} ignored.".format(tarball.name + ".license", word))
 
     content = read_conf_file(os.path.join(path, "golang_libpath"))
     if content and content[0]:
--- a/autospec/license.py
+++ b/autospec/license.py
@@ -67,8 +67,10 @@ def add_license(lic):
     real_lic_str = config.license_translations.get(lic, lic)
     real_lics = real_lic_str.split()
     for real_lic in real_lics:
-        if real_lic in licenses or real_lic in config.license_blacklist:
+        if real_lic in config.license_blacklist:
             continue
+        elif real_lic in licenses:
+            result = True
         else:
             result = True
             licenses.append(real_lic)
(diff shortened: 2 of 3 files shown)


fenrus75:
can we make it say non-SPDX

it's not so much ambiguous as not spdx

jurobystricky:
> can we make it say non-SPDX it's not so much ambiguous as not spdx

That is not quite true. If I put "MY_DUMMY_LICENSE" in abc.licence, it will be initially accepted. (The build will complain about non-SPDX license at the very end, though)

phmccarty:
@jurobystricky Autospec has no understanding about SPDX versus non-SPDX. Since you are making the warning/error reporting more fine-grained for blacklisted licenses, I think I would prefer something like:

```
[WARNING] lapack.license: blacklisted license identifier ignored: "GPL"
[WARNING] lapack.license: blacklisted license identifier ignored: "BSD"
```

phmccarty:
@jurobystricky Also note that the travis checks fail due to the extra blank line added. Please remove that line.

jurobystricky:
I will make the changes in the message (ambiguous -> blacklisted) . As for Travis, it also does not like a comma w/o a space in front of it.  (I don't think I added the blank line, but I will remove it,  I'll make sure the next patch passes Travis.)

jurobystricky:
A different unittest snag. I will look into that.

jurobystricky:
The unittest fails if adding a valid license via "add_license" returns True if the license is already  present. The test expects False in this case.

"add_license" changed behavior slightly due to the patch. The returned value from "add_license" is set to False only  if we want to add a license which is invalid (blacklisted). Previously it was also set to False if we wanted to add a valid license but the license was already present.  Therefore "add_license" could fail for both valid and invalid licenses.

With this patch "add_license" only fails with an invalid license. If  a valid license is already present "add_license" will return True, but without adding the license to the list.

I can roll back the behavior of "add_license".  I would need to revisit the patch in order to avoid a different set of potentially misleading messages. 

OTOH I am fairly positive the unittest is the only place that checks the return value of "add_license", so I could easily change the unittest code to reflect the modified behavior of "add_license".

bryteise:
I think updating the unit test to reflect the new behavior is the way to go. Also should verify that the license doesn't get added to the list again.

jurobystricky:
Third time's a charm. Travis is happy.

PR #459, opened 2019-09-03
maven pattern: do not install_append twice

phmccarty:
write_install_append() is already called by write_buildpattern(), so it should not be called in pattern-specific code.


bryteise:
That's reasonable.

phmccarty:
I initially discovered the duplicate `install_append` snippets when trying to re-autospec apache-maven, and this PR resolves it.

PR #461, opened 2019-09-03
Add support for mvn() virtual provides; fix up existing jar mappings

phmccarty:
A bunch of virtual provides were added to `mvn-*` packages, so adapt some current fail patterns to add BuildRequires for them.

Also, refresh existing jar mappings for other fail patterns that still need them.


phmccarty:
One additional improvement I am thinking about here is to add the pom buildreq associated with the jar when encountering the `[WARNING] The POM for ... is missing, no dependency information` fail pattern. Right now, it just adds a buildreq for the jar.

bryteise:
Seems reasonable, I really need a dump of what is going on in autospec tomorrow I think.

PR #462, opened 2019-09-04
Recursively search for target jar/pom files

bwarden:
Installs jar and pom files for any target directory, not just ./target.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -641,22 +641,40 @@ class Specfile(object):
         """Write installation steps to specfile for maven source packages."""
         self.write_license_files()
 
+        # Iterate over all target subdirs
+        self._write_strip("for targetdir in $(find . -type d -name target); do")
+        self._write_strip("pushd $targetdir")
+
         # Figure out the artifact details and path components
-        self._write_strip("export GROUP_PATH=$(xml sel -t -v '/_:project/_:groupId' pom.xml | sed 's#\\.#/#g')")
-        self._write_strip("export ARTIFACT_ID=$(xml sel -t -v '/_:project/_:artifactId' pom.xml)")
-        self._write_strip("export VERSION=$(xml sel -t -v '/_:project/_:version' pom.xml)")
+        # Find group ID; inherit from parent if necessary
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -641,22 +641,39 @@ def write_maven_install(self):
         """Write installation steps to specfile for maven source packages."""
         self.write_license_files()
 
+        # Iterate over all target subdirs
+        self._write_strip("for targetdir in `find . -type d -name target`; do")
```
phmccarty: For consistency in shell code style, I would prefer seeing `$(...)` here rather than the backticks.

autospec/specfiles.py
```diff
@@ -641,22 +641,39 @@ def write_maven_install(self):
...
+        self._write_strip("export VERSION=$(xml sel -t -v '/_:project/_:version' -v '/_:project/_:parent/_:version' ../pom.xml | head -1)")
 
         # Create the installation path
         self._write_strip("export DEPLOY_PATH=%{buildroot}/usr/share/java/.m2/repository/${GROUP_PATH}/${ARTIFACT_ID}/${VERSION}")
         self._write_strip("mkdir -p ${DEPLOY_PATH}")
 
         # Copy all the jar files
-        self._write_strip("cp -p target/${ARTIFACT_ID}*.jar ${DEPLOY_PATH}/")
+        self._write_strip("for jarfile in ${ARTIFACT_ID}*.jar; do")
+        self._write_strip('if [[ -f "${jarfile}" ]]; then')
```
phmccarty: This conditional is not needed, unless I am missing some detail.

autospec/specfiles.py
bwarden: Agreed.

autospec/specfiles.py
bwarden: If the glob doesn't match anything, $jarfile literally contains "*"

phmccarty: Ah, I didn't consider it would be possible for the glob to match nothing...

So if that is true, then we can either keep the conditional, or set `nullglob` beforehand.

bwarden: I forgot all about `nullglob`. Fixing.

Issue #463, opened 2019-09-04
Implicit excludes not working for /usr/include/*.h


phmccarty:
I am trying to add a header file (from /usr/include/) to -extras, but the appropriate %exclude is not added to the -dev subpackage to override the header file glob.

bryteise:
Should be resolved with #464

PR #464, opened 2019-09-05
Stop globbing header files

bryteise:
This change eliminates the ability for header files to change path
without rebuilding. This limitation is being added because of changes
to exclude (to remove files rather than exclude them) making it
difficult to move header content to extras.


bryteise:
Fixes #463

phmccarty:
@bryteise By the way, I see no evidence of this change impacting rebuilds due to version upgrades, because headers themselves tend to not have versions in their, only headers within subdirectories of `/usr/include`. If RPM supported globs like `/usr/include/**/*.h` and we used them, only then would we see this type of impact.

PR #470, opened 2019-09-11
Add gradle build pattern.

AthenasJimenez:
Add support gradle build tool.

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -143,7 +143,7 @@ def failed_pattern(line, pattern, verbose, buildtool=None):
                 must_restart += buildreq.add_buildreq(config.gems[s], cache=True)
             else:
                 print("Unknown ruby gem match", s)
-        elif buildtool == 'maven':
+        elif buildtool == 'maven' or buildtool == 'gradle':
             group_count = len(match.groups())
             if group_count == 2:
                 # Add fully qualified versioned mvn() dependency
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -307,7 +307,10 @@ failed_pats = [
     (r"Add the installation prefix of \"(.*)\" to CMAKE_PREFIX_PATH", 0, None),
     (r"^.*\"(.*)\" with any of the following names.*$", 0, None),
     (r"module not found: .*#(.*);(?:.*)", 0, 'maven'),
-    (r"-- artifact (?:.*)#(?:.*);(?:.*)!(.*).jar", 0, 'maven')]
+    (r"-- artifact (?:.*)#(?:.*);(?:.*)!(.*).jar", 0, 'maven'),
+    (r"\[WARNING\] Unable to get resource '(?:.*):(.*):[jar|pom]*:(?:.*)' from repository", 0, 'maven'),
+    (r"\> Could not find (?:.*):(.*):(?:.*)\.$", 0, 'maven'),
+    (r"\> Could not resolve (?:.*):(.*):(?:.*)\.$", 0, 'maven')]
 
 
 def get_metadata_conf():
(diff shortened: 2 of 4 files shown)


bryteise:
Looks like a reasonable start of things to me.

autospec/build.py
```diff
@@ -143,7 +143,7 @@ def failed_pattern(line, pattern, verbose, buildtool=None):
                 must_restart += buildreq.add_buildreq(config.gems[s], cache=True)
             else:
                 print("Unknown ruby gem match", s)
-        elif buildtool == 'maven':
+        elif buildtool == 'maven' or buildtool == 'gradle':
```
bryteise: So if the buildtool is gradle it will also add mvn_provide, is that expected?

phmccarty: Yes, that is expected, because we are installing all artifacts to the same repository tree, and every jar, pom, or signature artifact will have a `mvn()` virtual provide added for it.

phmccarty: On second thought, this change is a no-op, because `maven` is still specified for the new gradle fail patterns.

autospec/config.py
```diff
@@ -307,7 +307,10 @@
     (r"Add the installation prefix of \"(.*)\" to CMAKE_PREFIX_PATH", 0, None),
     (r"^.*\"(.*)\" with any of the following names.*$", 0, None),
     (r"module not found: .*#(.*);(?:.*)", 0, 'maven'),
-    (r"-- artifact (?:.*)#(?:.*);(?:.*)!(.*).jar", 0, 'maven')]
+    (r"-- artifact (?:.*)#(?:.*);(?:.*)!(.*).jar", 0, 'maven'),
+    (r"\[WARNING\] Unable to get resource '(?:.*):(.*):[jar|pom]*:(?:.*)' from repository", 0, 'maven'),
+    (r"\> Could not find (?:.*):(.*):(?:.*)\.$", 0, 'maven'),
+    (r"\> Could not resolve (?:.*):(.*):(?:.*)\.$", 0, 'maven')]
```
bwarden: We can get org and version out of these here, too, to generate a full `mvn({org}:{name}:jar) = {ver}`. I can submit a separate PR to get that.

autospec/specfiles.py
```diff
@@ -1586,6 +1586,19 @@ def write_ant_pattern(self):
...
         self._write_strip("")
 
+    def write_gradle_pattern(self):
+        self.write_prep()
+        self._write_strip("%build")
+        self.write_build_prepend()
+        self.write_proxy_exports()
+        self._write_strip("mkdir -p /builddir/.m2")
+        self._write_strip("cp -r /usr/share/java/.m2/* /builddir/.m2/")
+        self._write_strip("gradle --offline " + self.extra_make)
```
bwarden: Did you find a way to get gradle to use Maven's local repo by default? So far I only came up with this:
`sed -i 's/\(repositories\s*{\)/\1\n    mavenLocal()/' build.gradle`

autospec/build.py
AthenasJimenez: Yup, it would use all mvn pkgs just like any other maven packages.

autospec/config.py
AthenasJimenez: sounds good to me, please do :)

autospec/specfiles.py
AthenasJimenez: Sadly, no. Now I'm reading the documentation for newer releases of gradle to see if there is an improvement we could use. The offline flag only search on the gradle cache for the missing stuff.

autospec/config.py
phmccarty: Does gradle print the group id, artifact id, version, and (optional) classifier in the same way as maven? If not, then the new capture group handling in `failed_pattern()` will need further refactoring to support the gradle messages.

bwarden: `group:artifact:version`

I haven't seen classifier (or e.g. `pom`, `jar`, etc.)

PR #471, opened 2019-09-11
Profile payload refactor

phmccarty:
Move the duplicated code into a new function, and have all of the patterns use it if they support it. Also, add a stub for the `make` pattern.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -572,6 +572,48 @@ class Specfile(object):
                 file2 = file.replace("/", "_")
                 self._write_strip("cp " + file + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + "\n")
 
+    def write_profile_payload(self, pattern=None):
+        """Write the profile_payload specified for this package."""
+        if not config.profile_payload:
+            return
+        use_subdir = True
+        init = ""
+        post = ""
+        if pattern == "configure":
+            init = f"{self.get_profile_generate_flags()}" \
+                   f"%configure " \
+                   f"{self.disable_static} " \
+                   f"{config.extra_configure} " \
+                   f"{config.extra_configure64}"
+        elif pattern == "configure_ac":
+            init = f"{self.get_profile_generate_flags()}" \
+                   f"%reconfigure " \
+                   f"{self.disable_static} " \
+                   f"{config.extra_configure} " \
+                   f"{config.extra_configure64}"
+        elif pattern == "autogen":
+            init = f"{self.get_profile_generate_flags()}" \
+                   f"%autogen " \
+                   f"{self.disable_static} " \
+                   f"{config.extra_configure} " \
+                   f"{config.extra_configure64}"
+        elif pattern == "cmake":
+            use_subdir = False
+            init = f"{self.get_profile_generate_flags()}"
+            post = f"{self.get_profile_use_flags()}"
+        if use_subdir and self.subdir:
+            self._write_strip("pushd " + self.subdir)
+        if init:
+            self._write_strip(init)
+        self.write_make_line()
+        if use_subdir and self.subdir:
+            self._write_strip("popd")
+        self._write_strip("\n")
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -572,6 +573,55 @@ def write_license_files(self):
                 file2 = file.replace("/", "_")
                 self._write_strip("cp " + file + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + "\n")
 
+    def write_profile_payload(self, pattern=None):
+        """Write the profile_payload specified for this package."""
+        if not pattern:
+            return
+        if not config.profile_payload:
+            return
+        if config.profile_payload and not config.profile_payload[0]:
```
bryteise: extra check on config.profile_payload

autospec/specfiles.py
```diff
@@ -572,6 +573,55 @@ def write_license_files(self):
...
+                   config.extra_configure,
+                   config.extra_configure64)
+        elif pattern == "cmake":
+            use_subdir = False
+            init = "{0}".format(self.get_profile_generate_flags())
+            post = "{0}".format(self.get_profile_use_flags())
+        else:
+            print_warning(f"profile_payload not supported for {pattern} pattern")
+            return
+        if config.profile_payload and config.profile_payload[0]:
```
bryteise: extra check on config.profile_payload

autospec/specfiles.py
```diff
@@ -572,6 +573,55 @@ def write_license_files(self):
...
+            if use_subdir and self.subdir:
+                self._write_strip("pushd " + self.subdir)
+            self._write_strip(init)
```
bryteise: since you check for post, maybe check for init too.

autospec/specfiles.py
```diff
@@ -572,6 +573,55 @@ def write_license_files(self):
...
+                   self.get_profile_generate_flags(),
+                   self.disable_static,
+                   config.extra_configure,
+                   config.extra_configure64)
+        elif pattern == "cmake":
+            use_subdir = False
+            init = "{0}".format(self.get_profile_generate_flags())
+            post = "{0}".format(self.get_profile_use_flags())
+        else:
+            print_warning(f"profile_payload not supported for {pattern} pattern")
```
bryteise: maybe this should be a hard error for testing

bryteise:
Seems like a reasonable approach.

autospec/specfiles.py
phmccarty: fixed

autospec/specfiles.py
phmccarty: fixed

autospec/specfiles.py
phmccarty: I removed this block in v2

phmccarty:
Changes since the v1 PR:
- Do not special-case the `make` pattern. Instead, if `pattern` is not passed to `write_profile_payload()` (or is `None`), the default behavior is assumed and no special init/post steps are needed.
- Switch to concatenated formatted string literals instead of calling `format`

autospec/specfiles.py
phmccarty: fixed in v2

autospec/specfiles.py
```diff
@@ -572,6 +572,50 @@ def write_license_files(self):
                 file2 = file.replace("/", "_")
                 self._write_strip("cp " + file + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + "\n")
 
+    def write_profile_payload(self, pattern=None):
+        """Write the profile_payload specified for this package."""
+        if not config.profile_payload:
+            return
+        if not config.profile_payload[0]:
+            return
```
bryteise: if not config.profile_payload or not config.profile_payload[0] probably but up to you.

phmccarty: I will remove the second conditional, since we determined that it's not needed.

PR #473, opened 2019-09-13
Extend gradle build pattern

bwarden:
Building on @AthenasJimenez work, I'm extending the build pattern and dependency matching.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -738,8 +738,11 @@ class Specfile(object):
         if self.subdir:
             self._write_strip("pushd " + self.subdir)
 
-        # The install goal copies all the pieces to the build/install subdir
-        self._write_strip("gradle --offline install -d " + self.extra_make_install)
+        # User must provide the install goal in make_install_args
+        self._write_strip("gradle --offline " + self.extra_make_install)
+
+        # Install from a build/install directory, if it was created
+        self._write_strip("if [[ -d build/install ]]; then")
 
         # Enter the installation subdir, so we can touch up what gets installed
         self._write_strip("pushd build/install")
@@ -775,6 +778,9 @@ class Specfile(object):
 
         # Leave the build subdir
         self._write_strip("popd")
+
+        # Done processing build/install directory
+        self._write_strip("fi")
 
         if self.subdir:
             self._write_strip("popd")
@@ -1653,8 +1659,8 @@ class Specfile(object):
         # Opportunistically report detected dependencies
         self._write_strip("gradle --offline dependencies || :")
 
-        # Execute the build goal
-        self._write_strip("gradle --offline build " + self.extra_make)
+        # Execute the build goal -- but user must provide it in make_args
+        self._write_strip("gradle --offline " + self.extra_make)
 
         if self.subdir:
             self._write_strip("popd")
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -1654,7 +1654,7 @@ class Specfile(object):
 
         # Point to our local maven repo, first
         # style check does not like the escapes here
-        self._write_strip(r"sed -i 's|\(repositories\s*{\)|\1\n    mavenLocal()|' build.gradle")
+        self._write_strip(r"find . -type f '(' -name '*.gradle' -o -name '*.gradle.kts' ')' -exec sed -i 's|\(repositories\s*{\)|\1\n    mavenLocal()|' {} +")
 
         # Opportunistically report detected dependencies
         self._write_strip("gradle --offline dependencies || :")
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -731,6 +731,54 @@ class Specfile(object):
         # All done
         self._write_strip("done")
 
+    def write_gradle_install(self):
+        """Write installation steps to specfile for maven source packages."""
+        self.write_license_files()
+
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
+        # The install goal copies all the pieces to the build/install subdir
+        self._write_strip("gradle --offline install -d " + self.extra_make_install)
+
+        # Enter the installation subdir, so we can touch up what gets installed
+        self._write_strip("pushd build/install")
+
+        # Handle jarfiles in lib dir
+        # The install goal results in packaging not only the jars that were
+        # built, but also runtime deps that came from other packages. For every
+        # jar we find in the installed directory, we'll check whether there's a
+        # matching jar in the M2 repository provided by installed build deps.
+        # If so, we'll replace this copy with a symlink.
+        # TODO: call add_requires() with an appropriately-derived value so the
+        # symlink will be resolved at chroot creation.
+        self._write_strip("shopt -s nullglob")
+        self._write_strip("for lib_dir in */lib; do")
+        self._write_strip("pushd ${lib_dir}")
+        self._write_strip("for jarfile in *.jar; do")
+        self._write_strip("JARNAME=$(basename ${jarfile})")
+        self._write_strip('REALJAR=$(find /usr/share/java/.m2/repository -type f -name "${JARNAME}")')
+        self._write_strip("if ! [[ -z ${REALJAR} ]]; then")
+        self._write_strip(r'ln -sf ${REALJAR/#\/usr\/share\/java/..\/..}')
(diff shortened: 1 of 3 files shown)


bwarden:
DO NOT MERGE YET.

I can successfully build the test package now, but I'm still working on a valid install function.

bwarden:
Style check does not like the regex escape sequences in the sed command

bwarden:
Ready to consider for merging

autospec/specfiles.py
```diff
@@ -1583,17 +1623,30 @@ def write_ant_pattern(self):
...
+        # It's ok if this doesn't exist
+        self._write_strip("cp -r /usr/share/java/.m2 ~/.m2 || :")
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
+        # Point to our local maven repo, first
+        # style check does not like the escapes here
+        self._write_strip(r"sed -i 's|\(repositories\s*{\)|\1\n    mavenLocal()|' build.gradle")
+
+        self._write_strip("gradle --offline dependencies")
```
phmccarty: Under what conditions is the `dependencies` command supposed to exit with an error? I just tried running `gradle dependencies` in the kafka 2.3.0 source root as a test, and I encounter the error below? Anything I'm missing, or is this expected for now?

```
FAILURE: Build failed with an exception.

* Where:
Build file '/tmp/kafka-2.3.0-src/build.gradle' line: 153

* What went wrong:
A problem occurred evaluating root project 'kafka-2.3.0-src'.
> Failed to apply plugin [id 'org.gradle.signing']
   > Could not create an instance of type org.gradle.plugins.signing.SigningExtension_Decorated.
      > Invalid signature file digest for Manifest main attributes

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
```

autospec/specfiles.py
```diff
@@ -1583,17 +1623,30 @@ def write_ant_pattern(self):
...
+        self._write_strip("gradle --offline build" + self.extra_make)
```
phmccarty: Probably want an extra space after `build` in the string.

autospec/specfiles.py
```diff
@@ -730,6 +730,46 @@ def write_maven_install(self):
...
+
+        self._write_strip("gradle --offline install -d")
+
+        # Enter the installation subdir
+        self._write_strip("pushd build/install")
+
+        # Handle jarfiles in lib dir
+        self._write_strip("shopt -s nullglob")
+        self._write_strip("for lib_dir in */lib; do")
+        self._write_strip("echo BTW OK 2")
```
phmccarty: Looks like a debug command?

autospec/specfiles.py
```diff
@@ -730,6 +730,46 @@ def write_maven_install(self):
...
+        self._write_strip("pushd ${lib_dir}")
+        self._write_strip("for jarfile in *.jar; do")
+        self._write_strip("JARNAME=$(basename ${jarfile})")
+        self._write_strip('REALJAR=$(find /usr/share/java/.m2/repository -type f -name "${JARNAME}" | head -1)')
```
phmccarty: I'm trying to understand the ` | head -1` bit. If there is more than jar file with that name in the repository, how do we know we have chosen the correct one?

On the other hand, if there is only one possible match for the jar file name in the repository, then why is ` | head -1` needed?

autospec/specfiles.py
```diff
@@ -730,6 +730,46 @@ def write_maven_install(self):
...
+
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
+        self._write_strip("gradle --offline install -d")
+
+        # Enter the installation subdir
+        self._write_strip("pushd build/install")
+
+        # Handle jarfiles in lib dir
```
phmccarty: I think this code comment could be expanded to describe in greater detail the rationale behind the code below that conditionally creates the jar symlinks.

autospec/specfiles.py
bwarden: Hmmm, good question. I only stuck this in there to give us something potentially to scan; it's not needed for the build. We can take it out, or make it opportunistically non-fatal instead.

autospec/specfiles.py
bwarden: Oops, yes.

bwarden: I'll also add `self.extra_make_install` in the appropriate place.

autospec/specfiles.py
bwarden: Yes!

autospec/specfiles.py
bwarden: This was a leftover from an attempt where I searched multiple directories and wanted to prioritize the first result. But yeah, it would be better to fail here if we get multiples.

autospec/specfiles.py
bwarden: Adding comments

autospec/specfiles.py
```diff
@@ -1641,7 +1641,8 @@ def write_gradle_pattern(self):
             self._write_strip("pushd " + self.subdir)
 
         # Point to our local maven repo, first
-        self._write_strip("sed -i 's/\(repositories\s*{\)/\\1\\n    mavenLocal()/' build.gradle")
```
AthenasJimenez: We need to add other files like:

sed -i 's/\(repositories\s*{\)/\1\n mavenLocal()/' settings.gradle
sed -i 's/\(repositories\s*{\)/\1\n mavenLocal()/' build.gradle.kts
sed -i 's/\(repositories\s*{\)/\1\n mavenLocal()/' settings.gradle.kts
sed -i 's/\(repositories\s*{\)/\1\n mavenLocal()/' buildSrc/build.gradle.kts
sed -i 's/\(repositories\s*{\)/\1\n mavenLocal()/' buildSrc/settings.gradle.kts

And also, assume that all of them can fail, cause not all of them exists

bwarden:
Squashed

autospec/specfiles.py
bwarden: Is that a fixed list, or should I do something like:
`find . -type f '(' -name '*.gradle' -o -name '*.gradle.kts' ')' -exec sed -i ...`

autospec/specfiles.py
```diff
@@ -1583,17 +1629,29 @@ def write_ant_pattern(self):
...
+        # It's ok if this doesn't exist
+        self._write_strip("cp -r /usr/share/java/.m2 ~/.m2 || :")
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
+        # Point to our local maven repo, first
+        self._write_strip("sed -i 's/\(repositories\s*{\)/\\1\\n    mavenLocal()/' build.gradle")
+
+        self._write_strip("gradle --offline dependencies")
+        self._write_strip("gradle --offline build" + self.extra_make)
```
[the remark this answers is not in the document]
bwarden: OK. It seemed like a more reliable goal than with maven, but we don't have to hard-code that assumption.

autospec/specfiles.py
```diff
@@ -730,6 +730,52 @@ def write_maven_install(self):
...
         self._write_strip("done")
 
+    def write_gradle_install(self):
+        """Write installation steps to specfile for maven source packages."""
+        self.write_license_files()
+
+        if self.subdir:
+            self._write_strip("pushd " + self.subdir)
+
+        self._write_strip("gradle --offline install -d")
```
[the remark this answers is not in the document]
bwarden: Hmm. What I saw implied that gradle goals were a lot more dependable. If we can't count on it, though, we'll definitely have a lot of boilerplate in `install_append`.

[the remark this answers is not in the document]
bwarden: How about if I leave it to the user to specify the install goal, but afterward, if `build/install` exists, then descend into it and do the symlink and copy process I've implemented?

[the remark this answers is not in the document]
AthenasJimenez: ok, sounds good to me

autospec/specfiles.py
AthenasJimenez: thats a better option

autospec/specfiles.py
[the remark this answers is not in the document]
AthenasJimenez: great :)

Issue #477, opened 2019-09-18
autospec blindly changing giturl in options.conf


clsulliv:
Doing a 'make autospec' on ghostscript right now results in this change in options.conf:

-giturl = https://github.com/ArtifexSoftware/ghostpdl-downloads.git
+giturl = https://github.com/ArtifexSoftware/ghostscript.git

Unfortunately the old URL is more valid than the new URL, as the latter repo does not exist.

phmccarty:
Looks like autospec should not modify `giturl` value if it already set. Definitely a bug.

PR #479, opened 2019-09-18
Avoid overwriting giturl if it is already set

phmccarty:
When the source URL is from github.com, autospec was overwriting the
giturl in options.conf by mistake. It should only set a value if no
current value exists.

Fixes https://github.com/clearlinux/autospec/issues/477


bryteise:
Looks good to me.

PR #482, opened 2019-09-25
Fix bug with R description file parsing

phmccarty:
If `Imports:` or `Depends:` is immediately followed by a newline, then the regexp fails to match. See the R-psych package (version 1.8.12) for an example.

Fix the issue by adjusting the regexp to greedily match additional whitespace characters, as represented by the `\s` metacharacter.


phmccarty:
There are some additional parsing fixes that I want to integrate here, so I will wait until those changes are ready.

PR #484, opened 2019-09-27
Add 'dev_requires_staticdev' option to Autospec

gvancuts:
Add a new option called 'dev_requires_staticdev" that can be set via the package
`options.conf` file. Its purpose and behaviour is to make the `-staticdev`
subpackage be a required package by the `-dev` subpackage.

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -538,6 +538,10 @@ so_to_lib
   This option causes package ``.so`` files to be added to the ``lib`` subpackage
   instead of the ``dev`` subpackage.
 
+dev_requires_staticdev
+  If this option is set, the ``-staticdev`` package is required by the ``-dev``
+  package.
+
 autoupdate
   This option indicates that the package is trusted enough to be automatically
   update to its newest available version when set to ``true``. This flag is
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -127,6 +127,7 @@ config_options = {
     "security_sensitive": "set flags for security-sensitive builds",
     "so_to_lib": "add .so files to the lib package instead of dev",
     "dev_requires_extras": "dev package requires the extras to be installed",
+    "dev_requires_staticdev": "dev package requires the staticdev to be installed",
     "autoupdate": "this package is trusted enough to automatically update "
                   "(used by other tools)",
     "compat": "this package is a library compatability package and only "
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -248,6 +248,8 @@ class Specfile(object):
         deps["python"] = ["python3"]
         if config.config_opts['dev_requires_extras']:
             deps["dev"].append("extras")
+        if config.config_opts['dev_requires_staticdev']:
+            deps["dev"].append("staticdev")
         for k, v in self.custom_extras.items():
             if "requires" in v:
                 deps[k] = v['requires']


fenrus75:
where would this be used?

we need to try to limit use of static as much as possible

gvancuts:
> where would this be used?

I would use this for the `gnu-efi` package. The `devpkg-gnu-efi` is automatically created, but the bundle definition automatically created only includes the `-dev` package. However the static libraries (that were recently moved to a `-staticdev` subpackaged) are also needed else that devpkg cannot be used (to build ACRN for example).

> we need to try to limit use of static as much as possible

This is why I was thinking that instead of making this default, I would introduce this option which would be false for most of the packages and only the few ones that need it would set it to true.

fenrus75:
hmm but you're mixing autospec with bundles. if you want a devpkg-* lets
solve that in the bundle side not autospec side

gvancuts:
> hmm but you're mixing autospec with bundles. if you want a devpkg-* lets solve that in the bundle side not autospec side

That was my original idea but Auke pointed at me at autospec to resolve this via dependencies. That seemed like a viable option to me too but your proposed approach makes a lot of sense too.

I don't know if the code that does this automation (i.e. parsing the auto.devpkgs and generating the bundle files) is available somewhere, do you have a pointer? If that piece would also pick up the `-staticdev` in addition to the `-dev` package, it would solve my issue and I'm happy to create a patch for this.

ahkok:
This is a special case - gnu-efi ships `gnu-efi.pc` which says that it needs to link `-lgnu-efi`.

Normally, this means that `libgnu-efi.so` exists. In this package, only `libgnu-efi.a` exists.

In other words, one can conclude that if this situation happens, that the `-dev` subrpm should also include the `staticdev` subrpm automatically as a `Requires:`.

This doesn't seem to be a bundle issue. The issue happens when building `systemd` for instance, at the RPM level, which has nothing to do with bundles yet.

I don't want an option here, I prefer auto-detection of this case instead, which I think is a better solution.

gvancuts:
> In other words, one can conclude that if this situation happens, that the `-dev` subrpm should also include the `staticdev` subrpm automatically as a `Requires:`.

Do you know why a `-staticdev` subpackage was created in the first place? It did not exist before, it has caused a number of issues already but I do not know what the upside of having it is, do you?

> 
> I don't want an option here, I prefer auto-detection of this case instead, which I think is a better solution.

I agree that auto-detection would be preferred, let me look at how this could be done.

Are there other packages that you know of, besides `gnu-efi` that would benefit from such logic?

The lazyman approach could also be to drop the `auto.devpkgs` mechanism completely and maintain that one `devpkg-gnu-efi` bundle manually... this may be worth considering if no other packages would draw any benefit from this.

ahkok:
`.a` files are huge and not preferred at all. People should be linking dynamically. Only a very small set of use cases are legitimate reasons to link statically. And we want to avoid linking statically by accident.

`gnu-efi` is special because there is no dynamic loader in EFI. Everything needs to be static.

gvancuts:
> `.a` files are huge and not preferred at all. People should be linking dynamically. Only a very small set of use cases are legitimate reasons to link statically. And we want to avoid linking statically by accident.

That makes sense, thanks for the additional background.

> `gnu-efi` is special because there is no dynamic loader in EFI. Everything needs to be static.

And in this case, having a `-staticdev` subpackage does not make sense to me. Was that automatically created by switching to a newer version of autospec (and we need to resolve this corner case at the autospec level)? Or was there someone who purposedly created it, in which case I'm wondering what the intent behind it was.

The size of these two static libraries is not very big:
```
du -h /usr/lib64/libefi.a /usr/lib64/libgnuefi.a
756K	/usr/lib64/libefi.a
20K	/usr/lib64/libgnuefi.a
```

bryteise:
We added a staticdev subpackage to autospec causing the change for gnu-efi. This is a really one off kind of case to be honest and I'm not sure I want to teach autospec something weird like if a pkg-config file exists and there isn't any dynamic files in dev move the static files to dev (and I am really sure I don't want to do something like if there are no dynamic files in dev, move static files to dev).

I think I would take either this patch or have the things that require the static library to build add a buildreq_add and make the bundle change to distribute the staticdev.

fenrus75:
I'd prefer to avoid the autospec change to be honest; this is really a
total one off that we can cope with in other ways. If this were a common
case, sure.... but really it isn't and I'm not happy to do things
that make it too easy to do weird complicated static things

bryteise:
Totally reasonable to avoid this patch. I did like how it is pretty minimally intrusive and would enable people building packages against gnu-efi to not need to know how to handle this case. That said it is really one-off so special handling instead of teaching autospec weird things is quite nice too.

gvancuts:
So how about this?
```
diff --git a/auto.devpkgs b/auto.devpkgs
index ba0c6333..a8642d65 100644
--- a/auto.devpkgs
+++ b/auto.devpkgs
@@ -96,7 +96,6 @@ gnome-desktop-3.0
 gnome-settings-daemon
 gnome-software
 gnome-video-effects
-gnu-efi
 gnutls
 goa-1.0
 goa-backend-1.0
diff --git a/bundles/devpkg-gnu-efi b/bundles/devpkg-gnu-efi
index db4598bc..0886c8ac 100644
--- a/bundles/devpkg-gnu-efi
+++ b/bundles/devpkg-gnu-efi
@@ -8,3 +8,4 @@
 include(devpkg-base)
 
 gnu-efi-dev
+gnu-efi-staticdev

```

And we drop this patch obviously

bryteise:
@gvancuts Looks good to me!

gvancuts:
> @gvancuts Looks good to me!

Thanks, I've opened a PR for this then: https://github.com/clearlinux/clr-bundles/pull/156

PR #485, opened 2019-09-27
maven pattern: Improve the resolution for fail patterns

AthenasJimenez:
Adding the resolution for "parent" or "pom" artifacts when a
corresponding jar doesn't exists.

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -159,7 +159,10 @@ def failed_pattern(line, pattern, verbose, buildtool=None):
                 org = match.group(1)
                 name = match.group(2)
                 ver = match.group(3).replace('-', '.')
-                mvn_provide = f'mvn({org}:{name}:jar) = {ver}'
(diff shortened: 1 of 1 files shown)


autospec/build.py
```diff
@@ -159,7 +159,10 @@ def failed_pattern(line, pattern, verbose, buildtool=None):
                 org = match.group(1)
                 name = match.group(2)
                 ver = match.group(3).replace('-', '.')
-                mvn_provide = f'mvn({org}:{name}:jar) = {ver}'
+                if "-parent" in name or "-pom" in name or "-bom" in name:
```
phmccarty: Are these substrings suffixes? If so, I suggest to do:

```
if re.search("-(parent|pom|bom)$", name):
    ...
```

AthenasJimenez: Good suggestion ;)

PR #486, opened 2019-09-28
Introduce "build_append" and fix missing write_xxx_{prepend,append}

thiagomacieira:
Turns out there's a difference between build_append and install_prepend:
since they run from different shell scripts, environment variables set
at the beginning of %build will not be set in the snippets added by
install_prepend. Notably, that includes CFLAGS and SOURCE_DATE_EPOCH.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -78,6 +78,7 @@ class Specfile(object):
         self.cmake_srcdir = ".."
         self.prep_prepend = []
         self.build_prepend = []
+        self.build_append = []
         self.make_prepend = []
         self.install_prepend = []
         self.install_append = []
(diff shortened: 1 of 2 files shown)


autospec/specfiles.py
```diff
@@ -1378,7 +1392,9 @@ def write_ruby_pattern(self):
         self.write_build_prepend()
         self.write_proxy_exports()
         self._write_strip("export LANG=C.UTF-8")
+        self.write_make_prepend()
```
bryteise: I don't quite understand the addition of make_prepend use here for ruby.

autospec/specfiles.py
```diff
@@ -1529,7 +1545,9 @@ def write_cargo_pattern(self):
         self._write_strip("export http_proxy=http://127.0.0.1:9/")
         self._write_strip("export https_proxy=http://127.0.0.1:9/")
         self._write_strip("export no_proxy=localhost,127.0.0.1,0.0.0.0")
+        self.write_make_prepend()
```
bryteise: Similar to the ruby comment, this seems odd to me.

bryteise:
Hrm so general comment about the make_prepend, I think you are just adding in some missing calls as it improves consistency though I guess I hadn't really thought of their use outside of makefile-like builds.

thiagomacieira:
`make_prepend` is prepended to make, so it appears between the `%build` and the make-equivalent. It's important for native builds, since it's repeated for AVX2, AVX512 and 32-bit.

bryteise:
Hrm well no issue adding it for consistency though in the cases you added it, I don't think any support the additional instruction set builds.

PR #491, opened 2019-10-02
Support multiple versions of java packages

bwarden:
Implementation still underway; it's just getting big enough I would like some feedback before I get too far down the rabbit hole.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -831,8 +831,10 @@ def parse_config_files(path, bump, filemanager, version):
     # Parse the version-specific patch lists
     for version in versions:
         verpatches[version] = read_conf_file(os.path.join(path, '.'.join(['series', version])))
+        if any(p.lower().startswith('cve-') for p in verpatches[version]):
+            config_opts['security_sensitive'] = True
+            rewrite_config_opts(path)
 
-    # TODO - scan version-specific patches for CVEs
     if any(p.lower().startswith('cve-') for p in patches):
         config_opts['security_sensitive'] = True
         rewrite_config_opts(path)
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -90,6 +90,11 @@ cmake_modules = {}
 
 cves = []
 
+# Keep track of the package versions
+versions = OrderedDict()
+# Only parse the versions file once, and save the result for later
+parsed_versions = OrderedDict()
+
 # defines which files to rename and copy to autospec directory,
 # used in commitmessage.py
 transforms = {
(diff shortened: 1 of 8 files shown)


bwarden:
Currently, I handle the `versions` file, taking URLs from it. I generate the `SourceX:` entries in autospec and the corresponding `%setup` calls.

Next up: handling patches from the `series.<VER>` files.

autospec/tarball.py
```diff
@@ -143,24 +143,55 @@ def build_unzip(zip_path):
...
+            break
+
+    # Are there any files left?
+    if len(lines) < 1:
         print_fatal("Zip file doesn't appear to have any content")
         exit(1)
 
+    # Look for a common directory prefix
+    for line in lines:
+        if line[0] == '-':
```
bryteise: Under what condition does it hit this? I expect the header and footer being removed would cause no lines with '-' as the first character.

autospec/tarball.py
```diff
@@ -143,24 +143,55 @@ def build_unzip(zip_path):
...
+        filename = fields.pop()
+        if prefix is None:
+            prefix = filename
+        common = list()
+        for pair in zip(re.split(r'/', prefix), re.split(r'/', filename)):
+            if pair[0] == pair[1]:
+                common.append(pair[0])
+            else:
+                break;
+        prefix = '/'.join(common)
```
bryteise: If there is only 1 file in the zipfile (dumb but hey) isn't this going to have the prefix be the full path of that file?

bryteise: It also seems a little weird to me that prefix would potentially contain directories (ala src/foo).

autospec/tarball.py
```diff
@@ -143,24 +143,55 @@ def build_unzip(zip_path):
...
+
+    # If we didn't find a common prefix, make one, based on the zip filename
     if not prefix:
-        print_fatal("Malformed zipfile, unable to determine zipfile prefix")
-        exit(1)
+        zipfile = os.path.basename(zip_path)
+        prefix = os.path.splitext(zipfile)[0]
+        print("BTW: Faking prefix dir {}".format(prefix))
+        extract_cmd = "unzip -qq -d {0} {1}".format(
+                os.path.join(build.base_path, subdir), zip_path)
```
bryteise: subdir seems like it should be prefix here because I don't think it was ever defined.

bryteise: And fixed in the next commit X(. Well that's what I get for reviewing sequentially.

autospec/config.py
```diff
@@ -554,7 +559,17 @@ def parse_existing_spec(path, name):
...
-    return set(read_conf_file(os.path.join(path, "versions")))
+    versions = OrderedDict()
+    for line in read_conf_file(os.path.join(path, "versions")):
+        fields = line.split()
+        version = fields.pop(0)
+        if len(fields):
+            url = fields.pop(0)
+        else:
+            url = ""
+        versions[version] = url
```
bryteise: Hrm do we want to have some oops detection for somebody putting in multiple urls for the same version by accident?

autospec/tarball.py
bwarden: Whoops, relic to be deleted.

autospec/tarball.py
```diff
@@ -407,14 +407,15 @@ def set_multi_version(ver):
...
-            multi_version = set([ver])
+            multi_version = {ver:""}
     elif not multi_version:
         # Fall back to ensure a version is always set
         # (otherwise the last known version will be used)
-        multi_version = set("1")
-    return sorted(multi_version)[-1]
+        multi_version[1] = ""
+    latest = sorted(multi_version.keys())[-1]
+    return multi_version[latest]
```
bryteise: Maybe this changes in a later commit but I don't see why latest is being set here.

Does this change the return type to a url rather than a version? That seems off.

autospec/tarball.py
bwarden: Yes.
I would argue that we probably want it to be, as pretty much none of our known build systems first descend into a subdir.

autospec/tarball.py
bwarden: Already changed `prefix` to `subdir` in a subsequent commit and dropped the print. But I'm open to renaming `subdir` if ambiguous here.

autospec/config.py
bwarden: Currently any subsequent fields will be ignored when reading, and omitted when writing. Do you think we should report their existence? Could be done.

autospec/tarball.py
bryteise: And looking at the next patch this is resolved, oof this patch series is rough.

bwarden: Yeah, that was an oops. Later commit just returns `latest`.

autospec/tarball.py
bwarden: But if the pure view is that we should only be looking for a common, top-level directory automatically, I can rewrite to that.

autospec/config.py
bryteise: Yea, just a report warning so folks can see something.

autospec/tarball.py
bwarden: per conversation, starting with `os.path.dirname(filename)` instead of `filename` so we don't end up with the full path _including filename_ as the prefix when there's just one file.

bryteise:
Everything looks good to me at this point.

AthenasJimenez:
Took me a while, but everything looks good to me +1.

bwarden:
Alright, shall we merge this one, and I'll work on the remaining steps, while we watch to see whether this breaks other things?

bryteise:
We could though if you think the other changes are also risky to the rest of the ecosystem I think I'd rather have one potential for breakage rather than two atm.

bwarden:
Patches handled. Next up: multiple `%build` sections.

bwarden:
OK, this is now complete. It definitely needs to be tested on some well-understood builds to make sure it's doing the right thing, though.

bwarden:
Anything anyone else wants to see before we merge this?

AthenasJimenez:
As far as I'm concerned, no. Go ahead 👍

autospec/config.py
```diff
@@ -831,8 +831,10 @@ def write_default_conf_file(name, description):
     # Parse the version-specific patch lists
     for version in versions:
         verpatches[version] = read_conf_file(os.path.join(path, '.'.join(['series', version])))
+        if any(p.lower().startswith('cve-') for p in verpatches[version]):
+            config_opts['security_sensitive'] = True
+            rewrite_config_opts(path)
```
bryteise: Would you mind just setting a flag for rewrite_config to run outside the loop?

bwarden: Done

PR #493, opened 2019-10-03
maven pattern: Improve xml reading for version and groupId tags

AthenasJimenez:
Adding a if/else statement to avoid duplicity when reading version or
goupId tags (i.e. 3.2.03.2.0) from pom files.


AthenasJimenez:
@bwarden @phmccarty and @bryteise requesting your review manually cause I can't add it normally :(

PR #494, opened 2019-10-03
autospec: openmpi support

jurobystricky:
Initial support for concurrent serial and parallel build.
Serial build remains unaffected.

Created new setting in "options.conf" openmpi = false|true (false is default)

Support for additional build controlling files (equivalent to serial builds):
   "configure_openmpi"
   "cmake_args_openmpi"

Creation of additional packages:
   foo-openmpi
   foo-openmpi-dev
   foo-openmpi-staticdev

No support for AVX at present. The only build pattern fully supported is
"configure".

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -541,8 +541,8 @@ class Specfile(object):
             self._write_strip("export CC=clang\n")
             self._write_strip("export CXX=clang++\n")
             self._write_strip("export LD=ld.gold\n")
-            self._write_strip("CFLAGS=${CFLAGS/ -Wa,/ -fno-integrated-as -Wa,/}")
-            self._write_strip("CXXFLAGS=${CXXFLAGS/ -Wa,/ -fno-integrated-as -Wa,/}")
+            self._write_strip("CFLAGS=${CFLAGS/ -Wa,/ -fno-integrated-as -Wa,}")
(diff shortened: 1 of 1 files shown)


jurobystricky:
D'oh! I'll fix the Travis errors in a new PR.

jurobystricky:
I believe "cmake" pattern will also build/install correctly as-is if cmake called with 
<pre>
-DCMAKE_INSTALL_BINDIR=/usr/lib64/openmpi/bin
-DCMAKE_INSTALL_LIBDIR=/usr/lib64/openmpi/lib
-DCMAKE_INSTALL_INCLUDEDIR=/usr/include/openmpi-x86-64 
</pre>
I will verify this. If so, those definitions should be generated by autospec (or RPM/macro)

bryteise:
Hey sorry I didn't get to this one sooner. Would you mind rebasing and I'll take a look at it. Overall it seems pretty reasonable though.

jurobystricky:
I have some additional changes. I keep the latest code in my fork of the repo, currently rebased. It is probably better to have look there before I create a new PR, as it is still a bit of a moving target. I also rely on some new macros in clr-rpm-config (again, in my fork): %configure_openmpi, %cmake_openmpi. I successfully autospeced/built some packages, but there are some rough edges. Notably the absence of AVX support.

https://github.com/jurobystricky/autospec
https://github.com/jurobystricky/clr-rpm-config

jurobystricky:
The latest forced-pushed PR should be more complete and properly rebased.

PR #496, opened 2019-10-10
Improve gnome signature file acquisition

bryteise:


autospec/pkg_integrity.py
```diff
@@ -346,19 +346,24 @@ def fetch_shasum(shasum_url):
     @staticmethod
     def get_shasum(package_url):
         """Try and get sha url based on package url."""
-        for shasum_url in (package_url.replace(".tar.xz", ".sha256sum"),
-                           "{}.sha256sum".format(package_url)):
-            shasum = GnomeOrgVerifier.fetch_shasum(shasum_url)
-            if shasum:
-                return shasum
+        ext_pos = package_url.find(".tar.")
+        if not ext_pos:
```
phmccarty: `find` returns `-1` if the substring is not found, so I'm guessing this check should be different.

bryteise: Ah indeed, thanks!

PR #497, opened 2019-10-11
Fix detection of autoconf patches

bwarden:
A change for multi-version support, specifically to ensure that all
versions are scanned for licenses, broke the procedure that detects
whether to re-execute autoconf when patches affect input files. This
shifts the license scanning change into a more specific location to
leave the original autoconf scanning intact.


phmccarty:
+1, thanks

PR #498, opened 2019-10-14
Update prefixes dict if we rewrite URL

bwarden:
package URL is the key for the prefixes dict, so if we rewrite the URL,
we should make sure we have a corresponding entry in prefixes.


chivakker:
lgtm, tested where it was failing and now it is fixed.

PR #499, opened 2019-10-14
maven pattern: Improve jar files and pom files installation.

AthenasJimenez:
- Replace xmlstarlet with mvn to get artifact details. This would avoid
errors when reading pom files like empty values.

- Install components based on the pom files available in the project
instead of using target directories. This would avoid errors when pom
files create nested target directories or no one at all.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -748,17 +748,17 @@ class Specfile(object):
         for prefix in self.build_dirs.values():
             self._write_strip("cd ../{}\n".format(prefix))
 
-            # Iterate over all target subdirs
-            self._write_strip("for targetdir in $(find . -type d -name target); do")
-            self._write_strip("pushd $targetdir")
+            # Iterate over all directories with pom files
+            self._write_strip("for pomfile in $(find . -name pom.xml); do")
+            self._write_strip("pushd $(dirname ${pomfile})")
 
-            # Figure out the artifact details and path components
-            # Find group ID; inherit from parent if necessary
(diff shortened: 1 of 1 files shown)


AthenasJimenez:
@bryteise, @bwarden and @phmccarty  need you review guys :)

autospec/specfiles.py
```diff
@@ -748,17 +748,17 @@ def write_maven_install(self):
...
-            self._write_strip("export ARTIFACT_ID=$(xml sel -T -t -m '//_:project' -v '_:artifactId' ../pom.xml)")
-            # Find version -- this *might* be inherited from parent
-            self._write_strip("export VERSION=$(xml sel -T -t -m '//_:project' --if 'boolean(_:version)' -v '_:version' --else -v '_:parent/_:version' ../pom.xml)")
+            # Iterate over all directories with pom files
+            self._write_strip("for pomfile in $(find . -name pom.xml); do")
+            self._write_strip("pushd $(dirname ${pomfile})")
+
+            # Evaluate pom files to find artifact details and path components
+            # Find group ID
+            self._write_strip("export GROUP_PATH=$(mvn org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -Dexpression=project.groupId -q -DforceStdout | sed 's#\\.#/#g')")
```
bryteise: Do we need to have the plugin version listed as a static string?

AthenasJimenez: Yes, otherwise maven would ask for a different version depending on the project or on the current maven version. Changing plugin version could means that some parameters would not be available and we can end up with some errors :/

I know this is not pretty, so I'm open to suggestions :P

bryteise: I'm wondering if we could do some detection on the build system to get that version since in theory we'd have put the mvn-buildreq bits in place which would give access to that.

AthenasJimenez:
Closing since this won't be needed

PR #503, opened 2019-10-16
Add tests for banned paths

bryteise:
Autospec currently doesn't handle /etc/, /opt/, /usr/etc/, /usr/local/
or /usr/src/ content paths so explicity warn and then fail when
content in these paths are detected.

### DIFF
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -49,11 +49,31 @@ class FileManager(object):
         # /usr/include?  Yes in the general case, but for example for R
         # packages, the answer is No.
         self.want_dev_split = True
+        self.has_banned = False
+
+    @staticmethod
+    def banned_path(path):
+        """Check if the path is either banned or in a banned subdirectory."""
+        banned_paths = ["/etc",
+                        "/opt",
+                        "/usr/local",
+                        "/usr/etc",
+                        "/usr/src",
+                        "/var"]
+        for bpath in banned_paths:
+            if path.startswith(bpath):
+                return True
+        return False
 
     def push_package_file(self, filename, package="main"):
         """Add found %file and indicate to build module that we must restart the build."""
         if package not in self.packages:
             self.packages[package] = set()
+
+        if FileManager.banned_path(filename):
+            util.print_warning(f"  Content {filename} found in banned path, skipping")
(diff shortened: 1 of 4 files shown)


autospec/files.py
```diff
@@ -49,12 +49,31 @@ def __init__(self):
...
+        self.has_banned = False
+
+    @staticmethod
+    def banned_path(path):
+        """Check if the path is either banned or in a banned subdirectory."""
+        banned_paths = ["/etc",
+                        "/opt",
+                        "/usr/local",
+                        "/usr/etc",
+                        "/usr/src"]
```
phmccarty: Add `/var` to the list as well?

autospec/files.py
```diff
@@ -290,7 +309,6 @@ def push_file(self, filename):
             (r"^/etc/systemd/system/.*\.wants/", "active-units"),
             # now a set of catch-all rules
             (r"^/etc/", "config", "", "%config "),
```
phmccarty: Should these two patterns beginning with `/etc` be removed as well?

autospec/files.py
bryteise: Ah indeed, /var is something we run into now and then too thanks!

tests/test_build.py
```diff
@@ -424,6 +424,38 @@ def mock_util_call(cmd):
...
+                  'errors here\n'
+        m_open = mock_open(read_data=content)
+
+        with patch(open_name, m_open, create=True):
+            build.parse_build_results('testname', 0, fm)
+
+        build.util.call = call_backup
+
+        self.assertEqual(fm.has_banned, True)
+        # one for each file added
```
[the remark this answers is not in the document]
bryteise: Well we added 0 files so I wanted that to be captured. I could drop the comment but I don't really know what else to put there. Maybe `check no files were added`?

autospec/files.py
bryteise: Yes they should, thanks!

tests/test_build.py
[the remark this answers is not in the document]
phmccarty: yeah, that sounds fine

PR #504, opened 2019-10-17
Install /usr/lib/perl5 content to -data

phmccarty:
Because implicit requires of '/usr/bin/perl' exist for many packages,
perl-bin will be installed as a build dependency, but not the perl
package's modules; the modules currently live in the base package, and
-bin does not require that package.

The easiest solution for now is to ship the modules in -data, since -bin
requires -data. So, add a files pattern to ship all content matching
/usr/lib/perl5 in -data.


bryteise:
+1 I think we should consider doing python the same way at some point. Also I think having things shebang pull in dependencies or even just paths should get the base package rather than the specific subpackage, I imagine we'd be completely busted if something grabbed an extras subpackage either way though.

PR #509, opened 2019-10-28
Rework archives handling

bwarden:
Support archive files that don't include a top-level directory by making
one up. Also quit using %setup for archives since it causes interesting
side effects in %build that aren't really applicable for archives.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -444,10 +444,24 @@ class Specfile(object):
                     # Also JAR files
                     if archive.endswith('.jar'):
                         continue
-                    self._write_strip("cd ..")
-                    self._write_strip("%setup -q -T -D -n {0} -b {1}"
-                                      .format(prefix,
-                                              self.source_index[archive]))
+                    # Handle various archive types
+                    extract_cmd = 'tar xf {}'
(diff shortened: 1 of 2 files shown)


autospec/specfiles.py
```diff
@@ -444,10 +444,24 @@ def write_prep(self, ruby_pattern=False):
                     # Also JAR files
                     if archive.endswith('.jar'):
                         continue
-                    self._write_strip("cd ..")
-                    self._write_strip("%setup -q -T -D -n {0} -b {1}"
-                                      .format(prefix,
-                                              self.source_index[archive]))
+                    # Handle various archive types
+                    extract_cmd = 'tar xvf {}'
```
phmccarty: I think a non-verbose tar extract is preferable.

autospec/specfiles.py
```diff
@@ -444,10 +444,24 @@ def write_prep(self, ruby_pattern=False):
...
+                    if archive.endswith('.zip'):
+                        extract_cmd = 'unzip {}'
```
phmccarty: Similarly here, `unzip -q` is preferable.

PR #510, opened 2019-10-29
Fixes for CRAN description parsing

phmccarty:
Fixes some lingering bugs with the CRAN `DESCRIPTION` parsing, and also adds support for the `LinkingTo` keyword.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -314,8 +314,10 @@ def _get_desc_field(field, desc):
     create the list.
     """
     val = []
-    # Values may span multiple lines...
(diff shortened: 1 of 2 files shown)


phmccarty:
Would you like the examples in the form of unit tests? :)

bwarden:
Well sure :D, but I'm just thinking it's easier to read/verify regexes if there's an example right above of what they're trying to match.

phmccarty:
v2 series posted, with the following changes:
- Support the case when the target field is the last field in the file. None of the current 800+ CRAN packages need this, but better to be prepared for the future.
- Expand the parser function docstring with examples and clarifying explanation.
- Add several unit tests to cover much of the description parsing code.

PR #513, opened 2019-11-05
download.py: increase curl timeout to 60 seconds

jurobystricky:
Avoid errors such as:
"Unable to fetch xxx: Operation timed out after 29973 millisecond"

30 seconds may not be enough to download a large file
even with a fast connection or a smaller file with a slower
connection.


phmccarty:
Hmm, I thought that curl option was the `request timeout`, not how long it takes to download a file...

jurobystricky:
I repeatedly timed out after 30 seconds, trying to download 340MB archive using 100Mb connection. 
The transfer never entirely completed, due to the timeout. Bumping the timeout to 60 seconds fixed my problem, but bumping it more is certainly better, I don't see a downside.  
https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html

PR #520, opened 2019-11-11
Miscellaneous fixes

phmccarty:
- Fix a failing unit test. It needed an update with the recent changes to python test suite handling.
- Pass `--mock-opts` value for second mock call. (I did some cleanup to build.py in the second commit, so I recommend looking at the second and third commits separately.)

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -360,6 +360,7 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
         "--enable-plugin=ccache",
         f"--uniqueext={uniqueext}",
         cleanup_flag,
+        mockopts,
     ]
     ret = util.call(" ".join(cmd_args),
                     logfile=f"{download_path}/results/mock_build.log",
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -331,29 +331,49 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
         shutil.rmtree('{}/results'.format(download_path), ignore_errors=True)
         os.makedirs('{}/results'.format(download_path))
 
-    util.call("{} -r {} --buildsrpm --sources=./ --spec={}.spec "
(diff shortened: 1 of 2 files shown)


phmccarty:
BTW, the motivation for the `--mock-opts` fix is so that developers can run `make autospec MOCK_OPTS='\--old-chroot'`, which is especially important to have working if they are running autospec in a Docker container. I also had not used `MOCK_OPTS` in a long time so I never noticed that it was never honored for the second mock invocation.

autospec/build.py
```diff
@@ -331,29 +331,49 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
...
+        f"--root={mockconfig}",
+        "--buildsrpm",
+        "--sources=./",
+        f"--spec={tarball.name}.spec",
+        f"--uniqueext={uniqueext}",
+        "--result=results/",
+        cleanup_flag,
+        mockopts,
+    ]
+    util.call(" ".join(cmd_args),
```
bryteise: I'm a bit sad it has to be this way since call is just undoing the join but oh well. Maybe we could check the type arg for the command but meh.

autospec/build.py
```diff
@@ -360,6 +360,7 @@ def package(filemanager, mockconfig, mockopts, cleanup=False):
         "--enable-plugin=ccache",
         f"--uniqueext={uniqueext}",
         cleanup_flag,
+        mockopts,
```
bryteise: Should we have two different opts for srpm and brpm eventually? Well I'll wait for that until we have a use case.

phmccarty: Potentially, yes. I also would wait until we have a use case.

autospec/build.py
phmccarty: We could require `util.call()` to have a list for the first argument. That would be my preference.

PR #521, opened 2019-11-12
Refactoring tarball.build_unzip()

AthenasJimenez:
- Make use of zipfile library.
- Update test for this function adding multiple test cases.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -24,6 +24,7 @@ import os
 import re
 import shutil
 import subprocess
+import zipfile
 from collections import OrderedDict
 
 import build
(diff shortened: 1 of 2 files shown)


autospec/tarball.py
```diff
@@ -136,60 +137,36 @@ def build_unzip(zip_path):
...
-        filename = fields.pop()
-        if prefix is None:
-            prefix = os.path.dirname(filename)
-        common = list()
-        for pair in zip(re.split(r'/', prefix), re.split(r'/', filename)):
-            if pair[0] == pair[1]:
-                common.append(pair[0])
+    if zipfile.is_zipfile(zip_path):
+        with zipfile.ZipFile(zip_path, 'r') as content:
+            print(content.namelist())
```
bryteise: Looks like leftover debug print.

autospec/tarball.py
```diff
@@ -136,60 +137,36 @@ def build_unzip(zip_path):
...
+        print_fatal("Not a valid zip file.")
+        exit(1)
 
     # If we didn't find a common prefix, make a dir, based on the zip filename
     if not prefix:
-        subdir = os.path.splitext(zipfile)[0]
+        prefix = os.path.splitext(os.path.basename(zip_path))[0]
         extract_cmd = "unzip -qq -d {0} {1}".format(
-            os.path.join(build.base_path, subdir), zip_path)
+            os.path.join(build.base_path, prefix), zip_path)
```
bryteise: My memory isn't perfect here but the reason that subdir was used is because if prefix is returned when their wasn't actually a common prefix folder the files were in bad things happen. It would be nice if that could be fixed though. ping for @bwarden

autospec/tarball.py
AthenasJimenez: yeah, just notice. I'll remove it

autospec/tarball.py
AthenasJimenez: Mmm... I'll double check this one just to make sure.

autospec/tarball.py
```diff
@@ -136,60 +137,35 @@ def build_unzip(zip_path):
...
-        common = list()
-        for pair in zip(re.split(r'/', prefix), re.split(r'/', filename)):
-            if pair[0] == pair[1]:
-                common.append(pair[0])
+    if zipfile.is_zipfile(zip_path):
+        with zipfile.ZipFile(zip_path, 'r') as content:
+            lines = content.namelist()
+            # When zipfile is not empty
+            if len(lines) > 0:
+                # Take first argument as prefix, it usually is.
```
phmccarty: Should `lines` be sorted, in case the container directory is not first in the list?

AthenasJimenez: After reading os.path documentation, got a better solution. Replaced the code to identify and clean the prefix with prefix = os.path.commonpath(lines).

autospec/tarball.py
AthenasJimenez: Could not find a problem using prefix, maybe @bwarden can provide a specific example so I can try to reproduce it.

bwarden: I think mvn-xz was a good example of this, when attempting to build from source.

bwarden: https://tukaani.org/xz/xz-java-1.8.zip

AthenasJimenez: Yeah, just hit the issue. So, I'll return to subdir for now.

bryteise:
Looks good to me.

@AthenasJimenez Would you mind squashing these changes into a single commit? I think the original commit message makes sense as is.

AthenasJimenez:
Done @bryteise ;)

PR #522, opened 2019-11-13
Purge python2 support from autospec

bryteise:


autospec/files.py
```diff
@@ -254,8 +254,6 @@ def push_file(self, filename):
             (r"^/sbin/", "bin"),
             (r"^/bin/", "bin"),
             (r"^/usr/lib/python3.*/", "python3", "/usr/lib/python3*/*"),
-            (r"^/usr/lib/python2.*/", "legacypython", "/usr/lib/python2*/*"),
-            (r"^/usr/lib64/python.*/", "python", "/usr/lib64/python*/*"),
```
phmccarty: Was this mapping ever used?

bryteise: My memory is we originally only had a single mapping and so it was the default and we shortly there after needed both. I'd need to dive through the archives though to be sure.

PR #523, opened 2019-11-13
Enable container-based Action/Workflow for testing CI

phmccarty:
This commit sets up an in-tree, container-based Github Action and associated workflow for running `make check` and `make unittests` within a Clear Linux OS Docker container.

The most appropriate bundles to install in the container to satisfy all test dependencies are `python-extras` (includes `python3-basic` bundle and the `flake8` tool) and `package-builder` (includes `make`, and the `pycurl` package).

### DIFF
--- a/.github/actions/clearlinux-latest-action/Dockerfile
+++ b/.github/actions/clearlinux-latest-action/Dockerfile
@@ -1,0 +1,4 @@
+FROM clearlinux:latest
+RUN swupd bundle-add package-builder python-extras
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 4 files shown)


phmccarty:
This Github Action won't activated until/if the PR is merged, but you can see example test results in my fork:

https://github.com/phmccarty/autospec/commit/493420cec5ffe4eee88d54a302fc4e7b7dfd17c4/checks

bryteise:
Looks pretty clean.

bryteise:
Any documentation you could link for this or is all self discovery?

phmccarty:
The [hello world example](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action) is still the best resource I found.

With my first attempts at enabling the container-based workflow, I was installing the bundles via the workflow file (`tests.yml`) instead of just installing them in the Dockerfile, so obviously the container image was being generated repeatedly, once for each step in the workflow.

Once I made the connection that the Dockerfile is the correct place to the install the bundles to create a docker image for testing, I created a new step, "Create Docker Image", and then reuse that image in the next steps in the workflow.

I did not realize how simple it would be, because I wanted to make the bundle list configurable, but for autospec testing, we can just update the Dockerfile, since the action is private to this repo.

PR #524, opened 2019-11-14
Allow patch files as archives (for manual patching).

ahkok:


autospec/specfiles.py
```diff
@@ -444,6 +444,9 @@ def write_prep(self, ruby_pattern=False):
                     # Also JAR files
                     if archive.endswith('.jar'):
                         continue
+                    # Patch files
+                    if archive.endswith('.patch'):
+                        continue
```
bryteise: I assume the expected workflow is for patching archives or other things that can't be directly patched in sources?

ahkok: Yes, I needed to perform a patch at a time when %patch wouldn't work.

There also is the case of "we need to ship a patch file" for e.g. documentation purposes. We can `patch a patch` file into the tree, but this is cumbersome, and really difficult to read.

PR #531, opened 2019-11-20
Refactoring tarball.build_untar()

AthenasJimenez:
- Make use of tarfile library
- Update test for this function adding multipl test cases.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -24,6 +24,7 @@ import os
 import re
 import shutil
 import subprocess
+import tarfile
 import zipfile
 from collections import OrderedDict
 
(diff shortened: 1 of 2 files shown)


autospec/tarball.py
```diff
@@ -103,33 +104,27 @@ def check_or_get_file(upstream_url, tarfile, mode="w"):
...
-        file_type = subprocess.check_output(["file", tarball_path]).decode("utf-8").strip()
-        print_fatal("tarball inspection failed, unable to determine tarball contents:\n"
-                    "{}\n{}\n".format(file_type, cpe))
+    prefix = None
+    if tarfile.is_tarfile(tarball_path):
+        with tarfile.open(tarball_path, 'r') as content:
+            lines = content.getnames()
+            # When tarball is not empty
+            if len(lines) > 0:
+                prefix = os.path.commonpath(lines)
```
bryteise: This logic is fairly different from the original implementation. I would at least want test cases for the ./ being prepended to all the entries for instance (and then likely will need to fix the code to handle that as well). This isn't something that the zip format needed to worry about. As for the line not containing a directory, a tar file might be a single file and not a directory. In that case the prefix would be set to the filename which would have been an error before and I think we either want to fail in that case or at the very least not have the prefix be detected as the filename.

AthenasJimenez: Regarding to the cases with leading dot (./), added a test for that and commonpath() removes the leading dot from the dir name. I.e. If the commonpath is "./directory" it would only return "directory". My only concern, and not sure if it is a valid case, is if one of the elements of the list of files is "./", then it would assume there is no a common path.

On the other hand, I did not think about the case when tarball have a single file so yeah, I need to add a check up for that just o make sure and maybe adding a test case also.

bryteise: So it is a little odd to do but you can definitely make tar create an archive with:

```
./
./file1
./file2
```

As the contents. In this case it would be nice to get the zipfile handling behavior of subdir and that then also handles the case where the archive only contains files without a prefix (like a single file). We'd need to update specfile.py to handle that extraction similarly to zipfile though. Perhaps that's a change that can wait for another PR that tries to consolidate archive file processing (and figures out how to get rid of the subdir variable workaround for zipfile handling).

phmccarty: Regarding the case that @bryteise described in the previous comment, you can easily create a tarball like that for testing by running `tar -cf foo.tar .` in a directory containing some files.

AthenasJimenez: Used --one-top-level to create the directory and put the content in there, seems to work so far. :)

I would like to add that while I was testing the changes, found a curious situation for package php-APCu. Currently, autospec assumes the prefix is "apcu-5.1.18", but with this changes it cannot find a prefix. This is because the files list is [package.xml, apcu-5.1.18/, ..]

So, it creates a new directory and the structure ends being something like this:

apcu-5.1.18/
                    | - package.xml
                    | - apcu-5.1.18/
                                            | - file1
                                            | - file 2

This behavior is expected, cause not all the content is inside of a common directory. Just curious.

AthenasJimenez:
Yes, is a potential issue but I think we would cross that bridge when we come to it. Actually, I'm considering to remove build_un7z() from tarball, could not find a package that have a source with that compression.

autospec/tarball.py
bryteise: Ah, I was wondering if we had any packages doing this kind of thing or if it was a made up issue in my head. I like the layout with your changes.

bryteise:
You have a make check failure, mind rebasing on the lastest autospec, fixing the make check issue and repushing? Looks good otherwise.

phmccarty:
Oops, I didn't notice that `make check` failed with:

```
autospec/tarball.py:125:14: E126 continuation line over-indented for hanging indent
```

Otherwise, looks good to me. Rebasing the branch on latest master should fix `make unittests` as well.

PR #532, opened 2019-11-22
Remove capability for dynamic reserving of buildroots

phmccarty:
Instead of autospec reserving buildroots on-the-fly to not collide with buildroots that already exist, I would prefer for the developer (or automation) running autospec to explicitly modify mock's `basedir` config option whenever more control is needed, possibly combined with other config options that modify state locations (like `cache_topdir`, `root`, etc).

Considering only the `basedir` option, a developer could run these three commands simultaneously to build the package `foo` and also avoid buildroot collisions:
```
$ make autospec MOCK_OPTS="--config-opts=basedir=/var/lib/mock/dir1"
$ make autospec MOCK_OPTS="--config-opts=basedir=/var/lib/mock/dir2"
$ make autospec MOCK_OPTS="--config-opts=basedir=/var/lib/mock/dir3"
```
The build root names would be (assuming that `config_opts['root']='clear'`, as it is by default):
```
/var/lib/mock/dir1/clear-foo
/var/lib/mock/dir2/clear-foo
/var/lib/mock/dir3/clear-foo
```
(Note that a side-effect of this change is that all autospec unit tests now pass in Github Actions; tests were failing because `sudo` was not installed by default, and the unit tests required `sudo` via the `reserve_path()` function.)

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -22,7 +22,6 @@
 import os
 import re
 import shutil
-import subprocess
 
 import buildreq
 import config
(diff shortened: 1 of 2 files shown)


phmccarty:
Note that elsewhere, `/var/lib/mock/clear-<uniqueext>` is hardcoded, but I think I would like to solve those issues in the different PR, since the changes will be more invasive.

PR #536, opened 2019-11-26
license_translation: CPL->EPL-1.0

jurobystricky:
If CPL license is found by autospec, we end up with an
error:
ERROR: License 'CPL' is not an allowed SPDX license ID

CPL license was superseded by Eclipse Public License:
https://spdx.org/licenses/CPL-1.0.html

### DIFF
--- a/autospec/license_translations
+++ b/autospec/license_translations
@@ -23,6 +23,7 @@ BSD_2_clause, BSD-2-Clause
 BSD_3_clause, BSD-3-Clause
 Boost, BSL-1.0
 CC0, CC0-1.0
+CPL, CPL-1.0
 Expat, MIT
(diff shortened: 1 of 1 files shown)


jurobystricky:
One way or another, CPL needs to be translated, else autospec refuses to build with CPL.
The offending package (ADOL-C) actually specified explicitly Eclipse Public license 1.0, but autospec also found "CPL" within the ADOL-C included spec file. Hence the proposed translation from CPL->EPL-1.0. But I agree that translating CPL->CPL-1.0 is a better idea.

phmccarty:
Yeah, I think a `CPL` to `CPL-1.0` translation is appropriate.

PR #539, opened 2019-12-03
Read _append and _prepend scripts almost verbatim

thiagomacieira:
Don't skip empty lines or lines starting with # or simplify
whitespace. This makes non-trivial scripts nicer to read in the .spec
file, like those found in llvm.spec:

        for f in *; do
            ...indented lines go here...
            mv $f $f-$VERSION
            ln -s -f $f-$VERSION $f
        done

We skip the first line if it is a shebang or a file format annotation.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -458,9 +458,19 @@ def rewrite_config_opts(path):
     write_config(config_f, path)
 
 
-def filter_blanks(lines):
-    """Filter out blank lines from the line list."""
-    return [l.strip() for l in lines if not l.strip().startswith("#") and l.split()]
+def read_file(path, track=True):
+    """Read full file at path.
+
+    If the file does not exist (or is not expected to exist)
+    in the package git repo, specify 'track=False'.
+    """
+    try:
+        with open(path, "r") as f:
+            if track:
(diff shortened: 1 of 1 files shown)


bryteise:
So I think this is fine but I think I'd want to never have a heredoc as part of a script largely for reasons of it seems like a way to avoid adding a file as a patch. Generally we prefer (outside of a set of specific file types) to have files be added as part of a patch. Again your change isn't doing anything special aside from enabling shebang scripts to also work but I just wanted to comment on your reasoning. I am also not sure I'd bother with skipping anything at all as shebangs/modelines aren't problematic to being executed in the context of rpmbuild as far as I know (just acts like a comment).

thiagomacieira:
I don't want heredocs either (see #540). I can update the motivation for other reasons, it just makes the output nicer for some more-complex scripts (see llvm.spec).

bryteise:
I would be very happy to get an update to the motivating reason. Otherwise the change is fine for me.

phmccarty:
LGTM, though I agree with @bryteise; I would like to see an updated commit description that describes an example other than on-the-fly script creation via heredoc.

thiagomacieira:
Updated description.

PR #541, opened 2019-12-03
Special-case "kxmlgui" package: don't make it auto-require buildreq-kde

### LINKED ISSUE #487: buildreq_ban not applying for kxmlgui
```
$ cat Makefile
PKG_NAME := kxmlgui
URL = https://download.kde.org/stable/frameworks/5.62/kxmlgui-5.62.0.tar.xz
ARCHIVES = 

include ../common/Makefile.common
$ cat buildreq_ban
# This file contains build requirements that get picked up but are
# undesirable. One entry per line, no whitespace.
buildreq-kde
```

But Autospec always adds the dependency, causing the build to fail since that makes kxmlgui depend on kxmlgui. Added from:

```python
    if "kde.org" in url or "https://github.com/KDE" in url:                                                                                     
        buildreq.add_buildreq("buildreq-kde")

### PR DESCRIPTION
**Pull request by thiagomacieira**
Fixes #487, by way of a hacky workaround.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -471,7 +471,8 @@ def name_and_version(name_arg, version_arg, filemanager):
         buildreq.add_buildreq("buildreq-gnome")
 
     if "kde.org" in url or "https://github.com/KDE" in url:
-        buildreq.add_buildreq("buildreq-kde")
+        if "kxmlgui" not in url:
+            buildreq.add_buildreq("buildreq-kde")
 
     # SQLite tarballs use 7 digit versions, e.g 3290000 = 3.29.0, 3081002 = 3.8.10.2
     if "sqlite.org" in url:

### CODE REVIEW

File: autospec/tarball.py [hunk 1]
```diff
@@ -471,7 +471,8 @@ def name_and_version(name_arg, version_arg, filemanager):
         buildreq.add_buildreq("buildreq-gnome")
 
     if "kde.org" in url or "https://github.com/KDE" in url:
-        buildreq.add_buildreq("buildreq-kde")
+        if "kxmlgui" not in url:
+            buildreq.add_buildreq("buildreq-kde")
```
bryteise: I assume there is some ordering issue preventing kxmlgui from buildreq_banning buildreq-kde?

thiagomacieira: Probably, I don't know why buildreq_ban does not apply.

phmccarty: It is an ordering issue... a PR will be opened shortly.

bryteise: With @phmccarty change merged I'm closing this as buildreq_ban will apply to urls now too.

thiagomacieira:
Thanks.

PR #544, opened 2019-12-10
Update config.py

hnanni:
prevent extra \n being inserted in the spec file

example:
https://gist.github.com/hnanni/45eaa81e7a42b982611b85390e743c38

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -935,7 +935,7 @@ def parse_config_files(path, bump, filemanager, version):
 
     content = read_script_file(os.path.join(path, "make_check_command"))
     if content:
-        check.tests_config = '\n'.join(content)
+        check.tests_config = ''.join(content)
 
     content = read_conf_file(os.path.join(path, tarball.name + ".license"))
     if content and content[0]:


phmccarty:
@hnanni The extra newlines are being added as a consequence of how the new `read_script_file` function works. So, the `make_check_command` contents are not the only bits affected; `profile_payload` and other control files are affected, too.

To keep the changes to a minimum, I think I would prefer to modify `read_script_file` to remove trailing whitespace (including the newlines), and preserve leading whitespace.

phmccarty:
To achieve this, `read_script_file` could return:
```
[line.rstrip() for line in lines]
```

hnanni:
most of the files scaned with `read_script_file` just consume its result [e.g.](https://github.com/clearlinux/autospec/blob/8d3e09de2f90525872ac6eb32b4bd7177dfdc2a6/autospec/config.py#L976) 

```
$ grep profile_payload autospec/config.py
profile_payload = None
    global profile_payload
    profile_payload = read_script_file(os.path.join(path, "profile_payload"))
```
it is just `make_check_command` the one that got the ['\n'.join()](https://github.com/clearlinux/autospec/blob/8d3e09de2f90525872ac6eb32b4bd7177dfdc2a6/autospec/config.py#L938)

phmccarty:
Most of the `'\n'.join()` calls happen in autospec/specfiles.py

phmccarty:
For profile_payload, the spec file write happens in autospec/specfiles.py, line 699.

PR #551, opened 2019-12-16
fail patterns: replace kcalcore with kcalendarcore

phmccarty:
Upstream project has been renamed, so autospec fail patterns need updating.


phmccarty:
The package replacement took effect in build `31930`, so I will merge.

PR #558, opened 2020-01-07
test_build: Increase coverage for build errors patterns

AthenasJimenez:

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,7 +1,7 @@
 import unittest
 import tempfile
 import os
-from unittest.mock import patch, mock_open
+from unittest.mock import patch, mock_open, MagicMock
 import build
 import files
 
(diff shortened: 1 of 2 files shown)


tests/test_build.py
```diff
@@ -367,28 +367,24 @@ def mock_util_call(cmd):
...
+            builderrors = f.readlines()
+            for error in builderrors:
+                if not error.startswith('#'):
+                    input, output = error.strip('\n').split('|')
+                    build.buildreq.buildreqs = set()
+                    build.util.open_auto = mock_open(read_data=input)
+                    build.parse_build_results('testname', 0, fm)
+
+                    self.assertIn(output, build.buildreq.buildreqs)
+                    self.assertGreater(build.must_restart, 0)
```
bryteise: I am a bit worried that build.must_restart is 0 since a buildreq was added. Am I missing something?

phmccarty:
There are several new unit test failures. Can you take a look at those?

tests/test_build.py
phmccarty: It's an `assertGreater`, so it shouldn't be 0 at this point, right?

phmccarty:
@AthenasJimenez I suspect the other test failures are raised because `build.util.call` needs to be restored at the end of the test modified in this PR.

AthenasJimenez:
first try didn't work, so I'll keep looking :/

AthenasJimenez:
Fixed :)

bryteise:
Looks good to me.

tests/test_build.py
bryteise: Indeed!

PR #559, opened 2020-01-08
phpize pattern: attempt to "fix"

ahkok:
Commit 798d2a6b0d40583efe5b1c3a6413f605dde8bf73 introduced an unwanted
error for packages that are the exception and have files in the "rootless"
part of the tarball, but also a srcdir. This includes ALL the php modules.

In these cases, the CWD is changed for all the macros to run one level
deeper. We can account for that, easily, in autospec, and that is what
this change does.

However, for some reason the abireport list file ends up in that deeper
folder as well, and the RPM macros do NOT pick up the proper location,
causing RPM builds to still fail on this change. No idea how to solve
that part just yet.

### DIFF
--- a/tests/builderrors
+++ b/tests/builderrors
@@ -1,0 +1,118 @@
+    !  swig is not installed|perl(swig)
+    ExtUtils::Depends not installed|perl(ExtUtils::Depends)
+  Could not find swig|swig
+  Did not find swig|swig
+ exec: swig: not found|swig
+ mate-common.m4 not found|mate-common-dev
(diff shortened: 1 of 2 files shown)


AthenasJimenez:
Wondering, can we simply remove/ignore package.xml file from file list on tarball.py:build_untar()?

ahkok:
Yes, that may work as well?!

AthenasJimenez:
Yup, something like this should work: https://gist.github.com/AthenasJimenez/e9451790bf664bdc301fd60ad71af5e5

PR #560, opened 2020-01-09
config: Sort and consolidate fail patterns

AthenasJimenez:


bryteise:
This is great but could we do it in two patches because it is hard to review what changes were made for consolidation during the sort. So if you could sort in one patch and consolidate in another I'd greatly appreciate it.

autospec/config.py
```diff
@@ -225,40 +224,33 @@
     (r"CMake Error at cmake\/modules\/([a-zA-Z0-9]+).cmake", 0, None),
     (r"Can't locate [a-zA-Z0-9_\-\/\.]+ in @INC " r"\(you may need to install the ([a-zA-Z0-9_\-:]+) module\)", 0, 'perl'),
     (r"Cannot find ([a-zA-Z0-9\-_\.]*)", 1, None),
-    (r"Checking for (.*) development files... No", 0, None),
```
bryteise: What line is this rendered irrelevant by? I didn't see it sorry X(.

bryteise: Ah nevermind, spotted it on line 281.

phmccarty: line 281:
```
    (r"[Cc]hecking for (.*) (?:support|development files|with pkg-config)?\.\.\. [Nn]o", 0, None),
```

phmccarty: Haha, we collided :P

PR #563, opened 2020-01-11
Remove unresolvable names from buildreq_cache

### LINKED ISSUE #480: Unknown buildreqs should be removed from buildreq_cache
If an unknown buildreq is detected for a particular build round, autospec adds that buildreq to `buildreq_cache`. Autospec should avoid adding the unknown buildreq to `buildreq_cache` to avoid the need to manually remove it in later autospec runs.

phmccarty:
Fixes #480 

- Add a couple of helper functions to remove dependency names from the global sets
- Start using the helper functions for `buildreq_ban`, `pkgconfig_ban`, and `requires_ban`
- The root.log scan may reveal some unresolvable names, so remove those names from the sets as well.

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -191,7 +191,10 @@ def parse_buildroot_log(filename, returncode):
     for line in loglines:
         match = missing_pat.match(line)
         if match is not None:
-            util.print_fatal("Cannot resolve dependency name: {}".format(match.group(1)))
+            name = match.group(1)
+            util.print_fatal("Cannot resolve dependency name: {}".format(name))
+            # Avoid adding the dependency name to the buildreq cache
+            buildreq.remove_buildreq(name)
             is_clean = False
 
     return is_clean
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -77,6 +77,23 @@ def add_buildreq(req, cache=False):
     return new
 
 
+def remove_buildreq(req):
+    """Remove req from the global buildreqs and buildreqs_cache sets."""
+    global buildreqs
+    global buildreqs_cache
+
+    req = req.strip()
+
+    try:
+        buildreqs.remove(req)
+    except KeyError:
+        pass
+    try:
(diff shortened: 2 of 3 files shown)


autospec/build.py
```diff
@@ -191,7 +191,10 @@ def parse_buildroot_log(filename, returncode):
     for line in loglines:
         match = missing_pat.match(line)
         if match is not None:
-            util.print_fatal("Cannot resolve dependency name: {}".format(match.group(1)))
+            name = match.group(1)
+            util.print_fatal("Cannot resolve dependency name: {}".format(name))
+            # Avoid adding the dependency name to the buildreq cache
+            buildreq.remove_buildreq(name)
```
bryteise: Hrm so this removes it from the buildreq and rebuildreq_cache. Ideally we want to see if this is from the buildreq_cache or the buildreq_add and if it is in the cache just remove it and retry the build. If it is from the buildreq_add then it is fatal and the build should halt. Unfortunately we don't really keep the original file contents themselves around and just the consolidated sets.

phmccarty: Thanks, I had not considered the interaction between buildreq_add and buildreq_cache here..

So, the cases to handle are:

1. Unresolvable name in buildreq_add only -> halt build
2. Unresolvable name in buildreq_cache only -> eject from cache and start new build round
3. Unresolvable name in both files -> eject from cache and halt build

Does that look right?

phmccarty: > Unfortunately we don't really keep the original file contents themselves around and just the consolidated sets.

I didn't understand what you meant here at first, but after testing, it's very clear now :-P

I would like to solve that problem, too, but not sure how to go about it yet.

phmccarty:
I will split the last commit into a new PR, since that's a bug fix for #562

phmccarty:
I'm not entirely sure how I want to account for the gaps William mentioned. I will open a new PR at a later time once I determine a better solution.

PR #566, opened 2020-01-15
Read old version/patch info from git when possible

### LINKED ISSUE #433: CVE commit message missing if previous build was aborted
Autospec compares the spec file and series file to scan for CVE patches. If the previous build was aborted, those patches are already in the spec file and thus aren't used in the commit message. It may be better to compare against the spec file as last committed instead of the dirty copy for this purpose.

phmccarty:
In the process of updating packages to new releases, the build may fail,
and the spec file and other files are left in an intermediate,
uncommitted state.

This intermediate state is useful for some of the files, but for the
spec file (generally), we care more about the state transition from
commit to commit.

This change begins teaching autospec how to read spec file state from
the latest commit, specifically for the "old version" of the package and
the patch list from that old version.

Fixes #433

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -539,21 +539,48 @@ def parse_existing_spec(path, name):
     if not os.path.exists(spec):
         return
 
+    found_old_version = False
+    found_old_patches = False
+    ver_regex = r"^Version *: *(.*) *$"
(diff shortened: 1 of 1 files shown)


bryteise:
Ugh that's an ugly bug. LGTM.

PR #568, opened 2020-01-16
Ban a few percent-encoded characters

phmccarty:
These characters should never appear in version strings, because rpm
complains. If there are additional percent-encoded characters to add in
the future, we can update the list.


bryteise:
In the future it would be nice to only have strings that can't be in base64 encoded strings I think in that list.

Issue #567, opened 2020-01-15
Multi-signature verification support


phmccarty:
The release tarball for the current stable release of `gnupg`, version 2.2.19, was signed by two keys, producing a single detached signature file containing two concatenated signatures. However, autospec currently assumes that a sig file contains one signature.

Supporting multiple signatures will require some redesign of autospec's pkg_integrity module and how autospec tracks the keys, so I'm labeling this as an enhancement.

PR #569, opened 2020-01-17
Only verify the first GPG signature

phmccarty:
The only package in Clear Linux OS that (occasionally) has its source archive signed by multiple keys is `gnupg`, so to unblock updates for it, only verify the first signature from the .sig file.

When only one signature is present in the sig file (the common case), pass the entire sig file to `gpg --verify ...`, keeping the existing behavior.

For the implementation, I refactored `parse_key` into a more generic `parse_gpg_packets` function, and also updated tests.

Note that supporting verification of multiple signatures is much more involved, so I opened an enhancement request yesterday for that (#567).

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -123,10 +123,10 @@ class GPGCli(object):
             args = self.args + ['--dearmor', '--output', '-', signature]
             output, err, code = self.exec_cmd(args)
             if code != 0:
-                return None
+                return GPGCliStatus(f'Failed to convert {signature} to binary format')
             num_bytes = packets[0].get("length")
             if not num_bytes:
-                return None
+                return GPGCliStatus(f'Cannot verify first signature from {signature}')
             first_sig = output[:num_bytes]
             with tempfile.NamedTemporaryFile(prefix="newsig-", dir=self._home, delete=False) as new_sig_file:
                 new_sig_file.write(first_sig)
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -739,20 +739,28 @@ def parse_gpg_packets(filename, verbose=True):
             return None
         out = out.decode('utf-8')
         packets = []
+        packet = {}
         for line in out.splitlines():
+            complete = False
             m = re.search(r'^# off=(\d+) ctb=.* tag=.* hlen=(\d+) plen=(\d+).*$', line)
             if m:
-                packet = {}
                 packet["offset"] = int(m.group(1))
                 packet["length"] = int(m.group(2)) + int(m.group(3))
+            m = re.search(r'^:(signature) packet:.* keyid ([0-9A-F]+)$', line)
+            if m and "type" not in packet:
+                packet["type"] = m.group(1)
+                packet["keyid"] = m.group(2)
+                complete = True
+            m = re.search(r'^:(user ID) packet: "(.*) <(.+?)>"', line)
+            if m and "type" not in packet:
+                packet["type"] = m.group(1)
+                packet["user"] = m.group(2)
+                packet["email"] = m.group(3)
+                complete = True
+            # add the packet only if we've extracted all data we need from it
+            if complete:
                 packets.append(packet)
-            m = re.search(r'^:signature packet:.* keyid ([0-9A-F]+)$', line)
-            if m:
-                packets[-1]["keyid"] = m.group(1)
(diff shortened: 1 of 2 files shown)


phmccarty:
If this approach looks okay, I will look at adding an additional test for verification of the first GPG signature when multiple signatures are encountered.

phmccarty:
Hmm, I'm seeing [test errors](https://travis-ci.org/phmccarty/autospec/builds/638245154?utm_medium=notification&utm_source=email) for the travis CI in my fork... Interesting that those errors do not happen on Clear.

phmccarty:
The reason for test failure on Ubuntu is that "packet separator" lines (beginning with `# off=` are not present, and my changes rely on them being present.

bryteise:
Well hopefully the separator you are using isn't something they change to often.

autospec/pkg_integrity.py
```diff
@@ -714,33 +714,68 @@ def attempt_key_import(keyid, key_fullpath):
...
+                packet["length"] = int(m.group(2)) + int(m.group(3))
+                packets.append(packet)
+            m = re.search(r'^:signature packet:.* keyid ([0-9A-F]+)$', line)
+            if m:
+                packets[-1]["keyid"] = m.group(1)
+            m = re.search(r'^:user ID packet: "(.*) <(.+?)>"', line)
+            if m:
+                packets[-1]["user"] = m.group(1)
+                packets[-1]["email"] = m.group(2)
+        return packets
```
bryteise: I think I'd like this to be a little more resilient to unexpected data with hopefully some only minor changes.

Could we have a packet map defined outside the loop and then only append once the offset, length, keyid, user and email are found? Then if we find what looks to be a new packet before the old one is complete we just don't add that to the packets list. Thoughts?

autospec/pkg_integrity.py
```diff
@@ -114,7 +114,22 @@ def __init__(self, pubkey=None, home=None):
...
+        # first signature from the detached signature file.
+        sig_name = signature
+        packets = parse_gpg_packets(signature)
+        if len(packets) > 1:
+            # sig file may be ascii-armored, so dearmor it first...
+            args = self.args + ['--dearmor', '--output', '-', signature]
+            output, err, code = self.exec_cmd(args)
+            if code != 0:
+                return None
+            num_bytes = packets[0]["length"]
```
bryteise: Is it possible for length to not be defined here given how packets are currently added?

autospec/pkg_integrity.py
```diff
@@ -114,7 +114,22 @@ def __init__(self, pubkey=None, home=None):
...
+            first_sig = output[:num_bytes]
+            with tempfile.NamedTemporaryFile(prefix="newsig-", dir=self._home, delete=False) as new_sig_file:
+                new_sig_file.write(first_sig)
+                sig_name = new_sig_file.name
+        args = self.args + ['--verify', sig_name, tarfile]
```
bryteise: Do we want to do anything with the new_sig_file since it is generated like check it into the repo?

autospec/pkg_integrity.py
phmccarty: Not all packets will have all of those values set. But, yes, it would be more robust to initialize the packet list first, and then populate the packet info after we determine what data is present. I will make these changes.

autospec/pkg_integrity.py
phmccarty: Yes, good catch. It needs to check if the `length` key is present.

autospec/pkg_integrity.py
phmccarty: I'm not sure, hmm.

phmccarty: Given how rarely we see sig files containing multiple signatures, I would be inclined to not add the "new" sig file to the repo for now. At least its contents can always be derived from the signature file that *is* checked in.

bryteise: Sounds good, we should probably cleanup the temp file then at some point.

phmccarty: The temp file does get cleaned up (indirectly), because it is created in `self._home`, which is tmp directory that is cleaned up later.

phmccarty:
This series is ready for review again. I added two new tests for the multi-signature scenarios and fixed some error propagation issues in the `GPGCli.verify` method, thereby fixing the test for the error case when packet separator lines are not present in `gpg --list-packets`.

PR #572, opened 2020-01-30
tarball: Simplify Go packages management.

AthenasJimenez:
Add a special archives list (go_archives), so we can remove Go specific
methods and simply make use of regular methods.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -43,42 +43,16 @@ path = ""
 tarball_prefix = ""
 gcov_file = ""
 archives = []
+go_archives = []
 giturl = ""
 domain = ""
 prefixes = dict()
 
 
-def get_go_artifacts(url, target, ver):
-    """Get artifacts required to be a go proxy alternative."""
-    for name in [f"{ver}.{x}" for x in ["info", "mod", "zip"]]:
-        path = os.path.join(target, name)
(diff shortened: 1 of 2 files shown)


AthenasJimenez:
@bryteise would like to hear your feedback :P

autospec/tarball.py
```diff
@@ -530,6 +485,24 @@ def process_archives(archives):
             call(move_cmd)
 
 
+def process_go_archives():
+    """Set up extra archives required by go packages."""
+    global go_archives
```
bryteise: So I was really trying to avoid more globals in the other implementation. I realize this makes some things cleaner but I think globals tend to introduce more problems with understanding then it avoids once we reach a level of complexity (and tarbal.py is definitely there for me). It also potentially requires setting of global state between tests.

bryteise:
This is a really nice cleanup, with my one whine.

autospec/tarball.py
AthenasJimenez: Agree and tbh, I only used it as a temp solution while I'm finishing the design of a new class that would do a better management of sources. Anyway, I think I can find a way to remove the global, if really needed.

bryteise: I can live with this being intentionally temporary since you have a larger change in the works.

Issue #323, opened 2018-12-20
meson: test won't be reported in testresults file


chivakker:
Compiled igt-gpu-tools with autospec, The build finished successfully and the test command will do its job.  Although at the end testresults won't report properly

i.e. output from build.log

249/252 assembler: test/rndz                    OK       0.01 s
250/252 assembler: test/lzd                     OK       0.01 s
251/252 assembler: test/not                     OK       0.02 s
252/252 assembler: test/immediate               OK       0.01 s
Ok:                  246
Expected Fail:         6
Fail:                  0
Unexpected Pass:       0
Skipped:               0
Timeout:               0
Full log written to /builddir/build/BUILD/igt-gpu-tools-1.23/builddir/meson-logs/testlog.txt
+ exit 0

Unfortunately the output of testresults is

$ cat ./testresults 
Total : 0
Pass : 0
Fail : 0
Skip : 0
XFail : 0

There's a parsing problem introduced recently in autospec.

phmccarty:
I have a PR that will be ready soon that improves meson test result reporting.

PR #573, opened 2020-01-31
Implement meson tests autodetection and improve results reporting

### LINKED ISSUE #208: Auto-detect "check" command when using meson

phmccarty:
Fixes #208 and #323

### DIFF
--- a/autospec/check.py
+++ b/autospec/check.py
@@ -20,6 +20,7 @@
 #
 
 import os
+import re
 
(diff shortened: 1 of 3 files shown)


phmccarty:
Due to some inconsistencies in the code with the meson pattern handling of the `buildavx2` and `buildavx512` directories, I will omit the test execution in those directories for now.

PR #576, opened 2020-02-03
Use external tar instead of python's tarfile.

busykai:
tarfile module in python has limited support for compression.
Particularly, it does not support lzip (.lz). Since we're using tar to
ultimately extract the tarball, use it, instead of python's module, to
detect the contents too.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -22,7 +22,7 @@ import glob
 import os
 import re
 import shutil
-import tarfile
+import subprocess
 import zipfile
 from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


bryteise:
Ugh we just got subprocess out of that. I think ed is literally the only package using lzip at this point and the repo provides differently compressed tar files. Generally people use xz for lzma instead anyway so I am not particularly worried about this.

busykai:
OK. Withdrawn. I dislike relying on an incompatible implementation though, but I do understand potential problems with parsing a command output.

phmccarty:
@bryteise What repo are you referring to where differently compressed tars are available for `ed`? There's a new release that I'm trying to update to...

bryteise:
https://github.com/happy5214/gnu-ed/releases though obviously these aren't the release tars so maybe they won't work.

bryteise:
Might be able to send an email to the upstream mailing list and ask nicely though for non-lzip content.

PR #578, opened 2020-02-11
Stop adding maven build requirements

bryteise:
Turn off adding maven and gradle dependencies and build requirements
due to the reduction of support for this content in Clear Linux.


dorileo:
+1

PR #581, opened 2020-02-13
Consolidate source management in Source class.

AthenasJimenez:
This patch consolidates the code that manages the sources
(.zip, .tar, etc) in Source class. This class is written in a way
that allows the reuse of code for all current supported types of
sources and at the same time, it allows the easy addition of new
ones.

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -18,10 +18,8 @@
 #
 
 import configparser
-import glob
 import os
 import re
-import shutil
 import tarfile
 import zipfile
 from collections import OrderedDict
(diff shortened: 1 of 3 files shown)


autospec/tarball.py
```diff
@@ -43,12 +41,107 @@
...
+        else:
+            self.type = 'tar'
+
+    def get_prefix(self):
+        """Determine the prefix and subdir if no prefix."""
+        prefix_method = getattr(self, 'get_{}_prefix'.format(self.type))
+        prefix_method()
+        # When there is no prefix, create subdir
+        if not self.prefix:
+            self.subdir = os.path.splitext(os.path.basename(self.path))[0]
```
bryteise: I think I'd prefer the get_* to return a value. This is a set_*.

autospec/tarball.py
```diff
@@ -43,12 +41,107 @@
...
+                if len(lines) == 0:
+                    print_fatal("Tar file doesn't appear to have any content")
+                    exit(1)
+                elif len(lines) > 1:
+                    if 'package.xml' in lines and buildpattern.default_pattern in ['phpize']:
+                        lines.remove('package.xml')
+                    self.prefix = os.path.commonpath(lines)
+        else:
+            print_fatal("Not a valid tar file.")
+            exit(1)
```
bryteise: Similarly this and zip are set_*.

autospec/tarball.py
```diff
@@ -43,12 +41,107 @@
...
+                else:
+                    print_fatal("Zip file doesn't appear to have any content")
+                    exit(1)
+        else:
+            print_fatal("Not a valid zip file.")
+            exit(1)
+
+    def get_go_prefix(self):
+        """Set empty prefix for go packages (*.list)."""
+        self.prefix = ''
```
bryteise: This is also a set_*.

autospec/tarball.py
```diff
@@ -499,10 +471,62 @@ def process_go_archives():
         if url_mod not in archives:
             go_archives.extend([url_mod, ':'])
         if url_zip not in archives:
-            go_archives.extend([url_zip, f'{ver}/'])
+            go_archives.extend([url_zip, ''])
```
bryteise: Could you explain why you are dropping ver here?

tests/test_tarball.py
```diff
@@ -89,105 +89,30 @@ def test_version_configuration_override(self):
...
 
-        tarball.build_unzip.assert_called_once()
-        tarball.build_untar.assert_called_once()
+        tarball.Source.extract_zip.assert_called_once()
+        tarball.Source.extract_tar.assert_called_once()
 
-        tarball.build_unzip = build_unzip_backup
-        tarball.build_untar = build_untar_backup
+        tarball.Source.get_zip_prefix = get_zip_prefix_backup
+        tarball.Source.get_tar_prefix = get_tar_prefix_backup
```
bryteise: I think adding a test for ':' would make sense since that is a conditional branch. There are quite a few more conditionals that would also be nice to hit. I would like to get test coverage for your changes to be up near the 80% level.

autospec/tarball.py
```diff
@@ -43,12 +41,107 @@
...
+        self.path = path
+        self.type = None
+        self.prefix = None
+        self.subdir = None
+
+        # Extra  compressed archives
+        if not self.destination.startswith(':'):
+            self.get_type()
+            self.get_prefix()
+            self.extract()
```
bryteise: I think I'd like if extract wasn't done as part of creating the source object since this has an effect outside the object itself and instead somebody can get the created source object (optionally inspect it even) and then call extract but I'm not sold on that being useful so I am open to keeping it this way.

autospec/tarball.py
AthenasJimenez: Yes, basically because autospec don't use it.

Currently, it just add another dir:  v0.3.1/github.com/BurntSushi/toml@v0.3.1
With this patch: github.com/BurntSushi/toml@v0.3.1

It's important to notice that this is just for source exploration (finding dependencies, license, etc). At the end, it has no impact on the specfile so I have no problem to add it back.

tests/test_tarball.py
AthenasJimenez: Already working on adding the tests, but would take me some to finish them :/

autospec/tarball.py
AthenasJimenez: Mmm... think you are right. Maybe adding the extraction routine in inspection code would have more sense. I'll review that code and bring a new proposal.

autospec/tarball.py
bryteise: I think I was probably just being overly paranoid about conflicts though maybe it was due to a refactor making it no longer needed. I'm fine with dropping this considering the url uniqueness is reasonably certain.

tests/test_tarball.py
bryteise: This code being somewhat core to interacting with files, I'd like to ensure our coverage is healthy. I'm okay with seeing a set of tests that handle most of the scenarios (adding a case for archives/multiversion go seems helpful with these changes) rather than a strict coverage number.

autospec/tarball.py
bryteise: Alrighty, sounds good to me.

AthenasJimenez: Well, I ended up putting the extraction as part of tarball.process(). Still using Source methods, but not calling extract() as part of the object creation. If you are ok with this, I can also push the changes on test_tarball.py to fix the error and add coverage.

bryteise: Looks reasonable to me.

tests/test_tarball.py
AthenasJimenez: Just added more content with more scenarios. Most of the new content includes dynamic tests created on the fly, but also got some regular tests.

autospec/tarball.py
bwarden: Given that, would it make sense for this method to accept an optional value where the caller specifies the prefix to use?

bwarden: Or, maybe it should be find_*?

bryteise:
I really like the test changes, builds a nice framework to use for other types eventually.

autospec/tarball.py
```diff
@@ -499,10 +470,70 @@ def process_go_archives():
...
+        process_multiver_archives(main_src, multiver_archives)
+
+    full_archives = archives + go_archives + multiver_archives
+    # Download and extract full list
+    for arch_url, destination in zip(full_archives[::2], full_archives[1::2]):
+        src_path = check_or_get_file(arch_url, os.path.basename(arch_url), mode="a")
+        # Create source object and extract archive
+        archive = Source(arch_url, destination, src_path)
+        # Add archive prefix to list
+        buildpattern.archive_details[arch_url + "prefix"] = archive.prefix
```
bwarden: `archive.prefix` might be undefined here, should we test and possibly fall back to `archive.subdir`?

bryteise: So the code in specfiles tests if prefix isn't defined and uses essentially the subdir definition. Probably our change to cleanup that code litter in multiple places now, good catch.

PR #584, opened 2020-02-26
Use the pypi api for distutils3 builds

bryteise:
Get metadata about the package from pypi for distutils3 build patterns
then use that metadata to get summary, description and provides
information (license and requires maybe leveraged in the future).

The description isn't always very useful (often too long) so only use
it if there isn't one already.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -20,12 +20,15 @@
 #
 
 import ast
+import json
 import os
 import re
 import subprocess
 
 import buildpattern
 import config
+import download
+import specdescription
 import toml
 import util
 
@@ -37,6 +40,7 @@ requires = set()
 extra_cmake = set()
 verbose = False
 cargo_bin = False
+pypi_provides = None
 banned_buildreqs = set(["llvm-devel",
                         "gcj",
                         "pkgconfig(dnl)",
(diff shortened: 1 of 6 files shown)


autospec/buildreq.py
```diff
@@ -901,6 +906,27 @@ def scan_for_configure(dirn):
     else:
         config.autoreconf = False
 
+    if buildpattern.default_pattern == "distutils3":
+        # Try and grab the pypi details for the package
+        pypi_api_url = f"https://pypi.python.org/pypi/{tname}/{tversion}/json"
+        pypi_json = download.do_curl(pypi_api_url)
+        if pypi_json:
+            package_pypi = json.loads(pypi_json.getvalue().decode("utf-8"))
```
bwarden: Should we catch `JSONDecodeError` in case pypi gave us garbage?

autospec/buildreq.py
```diff
@@ -901,6 +906,27 @@ def scan_for_configure(dirn):
...
+            if package_pypi.get("info"):
+                info = package_pypi["info"]
+                if info.get("name"):
+                    pypi_provides = info["name"]
+                if info.get("requires_dist"):
+                    # Not yet adding requires
+                    pass
+                if info.get("license"):
+                    # The license field is freeform, might be worth adding though
```
bwarden: Maybe print it to the console as a suggestion.

bryteise: That seems reasonable enough to just do, thanks!

autospec/buildreq.py
bryteise: Seems like a safe thing to do.

PR #587, opened 2020-03-06
support for openmpi sections

jurobystricky:
Initial support for concurrent serial and parallel build.
Parallel MPI build enabled via a new setting in "options.conf":
    openmpi = false|true (false is default)

 * Serial builds remain unaffected.
 * Parallel build: Only OpenMPI is supported.
 * Parallel builds always assume AVX2.
 * Build patterns implemented: "configure" and "cmake".
   The build process is controlled by additional files:
   "cmake_arg_openmpi": same functionality as "cmake_args"
   "configure_openmpi": same functionality as "configure"

The openmpi (configure, build, check, install) sections follow these steps:

module load openmpi
...[configure,build,install,check]
module unload openmpi

Once "openmpi=true", the packages "openmpi-dev", "modules" and "openssh"
are automatically added to BuildRequires. ("openssh" is generally required
to run any MPI test suites).

Packages created:
   foo-openmpi:    binaries, libraries, docs, ...
   foo-dev :       contains both serial and MPI devel. files
   foo-staticdev:  contains both serial and MPI devel. files

### DIFF
--- a/autospec/files.py
+++ b/autospec/files.py
@@ -229,6 +229,7 @@ class FileManager(object):
         # the dev package. THis is useful for packages with a plugin
         # architecture like elfutils and mesa.
         so_dest = 'lib' if config.config_opts.get('so_to_lib') else 'dev'
+        so_dest_ompi = 'openmpi' if config.config_opts.get('so_to_lib') else 'dev'
 
         patterns = [
             # Patterns for matching files, format is a tuple as follows:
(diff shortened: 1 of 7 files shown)


autospec/files.py
```diff
@@ -241,6 +242,13 @@ def push_file(self, filename):
             (r"^/usr/share/info/", "info"),
             (r"^/usr/share/abi/", "abi"),
             (r"^/usr/share/omf", "main", "/usr/share/omf/*"),
+            (r"^/usr/lib64/openmpi/bin/", "openmpi"),
+            (r"^/usr/lib64/openmpi/share", "openmpi"),
+            (r"^/usr/lib64/openmpi/include/", "dev"),
+            (r"^/usr/lib64/openmpi/lib/[a-zA-Z0-9._+-]*\.so$", so_dest_ompi),
+            (r"^/usr/lib64/openmpi/lib/[a-zA-Z0-9._+-]*\.a$", "staticdev"),
+            (r"^/usr/lib64/openmpi/lib/[a-zA-Z0-9._+-]*\.so\.", "openmpi"),
+            (r"^/usr/lib64/openmpi/lib/", "dev"),
```
bryteise: So dev packages will not install the openmpi components, this means that if a user wants to develop against an openmpi library they'll need the corresponding openmpi bundle and the devpkg bundle where as normally the dev package would pull in the lib, bin components. This is a little bit broken since we expect a bundle to have all its dependencies to not have broken symlinks and the like.

jurobystricky: There is only a single  -dev package, containing also openmpi files, same file types as the non-openmpi (i.e. header files and *libs.so) . So the -dev package would include something like this:
<pre>
...
/usr/include/p4est.h
/usr/lib64/libsc-2.2.so
/usr/lib64/libsc.so
/usr/lib64/openmpi/include/p4est.h
/usr/lib64/openmpi/lib/libsc-2.2.so
/usr/lib64/openmpi/lib/libsc.so
...
</pre>
(For some reason github does not allow me to attach a file)

jurobystricky: I will make sure the -dev package will pull in -openmpi package (openmpi bin and libs are all in a single package)

bryteise: Great, thanks!

PR #588, opened 2020-03-09
Disable use of go modules

bryteise:
Unless a user specifically opts into using go modules don't try and
parse go.mod files for dependencies.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -825,16 +825,17 @@ def scan_for_configure(dirn, tname, dlpath):
                 # so only use go directly if we can't find a Makefile
                 buildpattern.set_build_pattern("golang", default_score)
             add_buildreq("buildreq-golang")
-            config.set_gopath = False
-            mod_path = os.path.join(dirpath, "go.mod")
-            reqs = parse_go_mod(mod_path)
-            for req in reqs:
-                # req[0] is a SCM url segment in the form, repo/XXX/dependency-name
-                # req[1] is the version of the dependency
-                pkg = "go-" + req[0].replace("/", "-")
-                add_buildreq(pkg)
-                if buildpattern.default_pattern == "godep":
-                    add_requires(pkg)
+            if buildpattern.default_pattern == "golang-mod" or buildpattern.default_pattern == "godep":
+                config.set_gopath = False
+                mod_path = os.path.join(dirpath, "go.mod")
+                reqs = parse_go_mod(mod_path)
+                for req in reqs:
+                    # req[0] is a SCM url segment in the form, repo/XXX/dependency-name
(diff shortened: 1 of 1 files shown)


autospec/buildreq.py
```diff
@@ -825,16 +825,17 @@ def scan_for_configure(dirn, tname, dlpath):
...
-            mod_path = os.path.join(dirpath, "go.mod")
-            reqs = parse_go_mod(mod_path)
-            for req in reqs:
-                # req[0] is a SCM url segment in the form, repo/XXX/dependency-name
-                # req[1] is the version of the dependency
-                pkg = "go-" + req[0].replace("/", "-")
-                add_buildreq(pkg)
-                if buildpattern.default_pattern == "godep":
-                    add_requires(pkg)
+            if buildpattern.default_pattern == "golong-mod":
```
phmccarty: the pattern name should be `golang-mod`, I think...

autospec/buildreq.py
```diff
@@ -825,16 +825,17 @@ def scan_for_configure(dirn, tname, dlpath):
...
+                config.set_gopath = False
+                mod_path = os.path.join(dirpath, "go.mod")
+                reqs = parse_go_mod(mod_path)
+                for req in reqs:
+                    # req[0] is a SCM url segment in the form, repo/XXX/dependency-name
+                    # req[1] is the version of the dependency
+                    pkg = "go-" + req[0].replace("/", "-")
+                    add_buildreq(pkg)
+                    if buildpattern.default_pattern == "godep":
```
phmccarty: This condition will never be true anymore, correct?

autospec/buildreq.py
```diff
@@ -825,16 +825,17 @@ def scan_for_configure(dirn, tname, dlpath):
                 # so only use go directly if we can't find a Makefile
                 buildpattern.set_build_pattern("golang", default_score)
```
phmccarty: Does this build pattern setting still make sense if we are checking for `golang-mod` below?

autospec/buildreq.py
bryteise: Oops, thanks!

autospec/buildreq.py
bryteise: Hrm right. I am not really worried about that condition for now but I will add another conditional in the top level that looks for "godep" as well as golang-mod.

autospec/buildreq.py
bryteise: I think in those cases we'd likely need to figure out a mechanism to build the package (but usually it is create a patch for the Makefile). For now it is probably the most sensible option but I'm open to suggestions.

phmccarty: Okay, let's leave it as-is for now.

PR #591, opened 2020-03-26
Fix bug with "file not found" detection

phmccarty:
Omitting the package NVR from the conditional will result in an
IndexError if the NVR does not actually exist on that line.

This change fixes the packaging for `epm` in Clear Linux OS, which
prints "file not found" errors when running its unit tests, but with
the i386 arch and slightly different NVR.


bryteise:
Ah nice.

PR #593, opened 2020-03-27
Refactor tarball to remove remaining globals

bryteise:
The tarball module had a number of globals that were referenced by
many other modules and has ordering dependencies with the config
module for some values. This made deciding on where certain values get
initialized difficult but before the initialization can be addressed a
refactor is helpful.

This change moves the global state (and functions that needed to
act on that global state) into a Content class. The goal of this work
is to better track what can be updated by a particular function, load
data in a sensible order and have it owned by a sensible component.

I expect that future work will see the tarball, buildpattern and some
portions of the config moved around to better match what component
should own what data (and to better deliniate between the data that
has a similar use but comes from different sources).

### DIFF
--- a/autospec/abireport.py
+++ b/autospec/abireport.py
@@ -27,7 +27,6 @@ import shutil
 import subprocess
 import sys
 
-import tarball
 import util
 
 valid_dirs = ["/usr/lib", "/usr/lib64"]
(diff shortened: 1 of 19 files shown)


bryteise:
I think there is a definite rework needed for most of these. Having random global functions on the modules is a bit weird. Still work for another time.

PR #595, opened 2020-03-31
tarball: update Git URL for GNOME projects

rossburton:
github.com/GNOME is a mirror of the canonical git repositories, which are at gitlab.gnome.org.


bryteise:
Thanks!

PR #596, opened 2020-04-01
Refactor build to remove remaining globals

bryteise:
The build module had a number of globals that were referenced by many
other modules and has ordering dependencies with the config and
tarball module for some values. This made deciding on where certain
values get initialized difficult but before the initialization can be
addressed a refactor is helpful.

This change moves the global state (and functions that needed to act
on that global state) into a Build class. The goal of this work is
to better track what can be updated by a particular function, load
data in a sensible order and have it owned by a sensible component.

Some of this data is likely to be removed from the build class in the
future as initialization of download and base paths should take place
in other modules most likely.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -60,7 +60,7 @@ def add_sources(download_path, archives, content):
     # /run or /tmp.
     #
     if os.path.exists(os.path.normpath(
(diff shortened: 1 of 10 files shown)


bryteise:
This is the last major refactor PR. Future changes are going to be reworking the data organization and consolidating ownership and functionality for the modules.

PR #597, opened 2020-04-02
Restore sorted order for pypi Requires lines

phmccarty:
In the recently removed `load_specfile()` function in buildreq.py (see
commit 2a181a2d), the pypi_requires set was sorted and converted to a
list via `sorted()`. Now, the set is no longer converted to a list, so
we can instead sort it immediately before writing out the contents in
`write_files_header()` in specfiles.py.


bryteise:
Nice catch!

Issue #580, opened 2020-02-13
AVX2/AVX512 instructions not generated for Fortran


jurobystricky:
When selecting "use_avx2 = true" and/or "use_avx512 = true" in options.conf,
autospec generates something like this:
<pre>
%build
...
export CFLAGS="$CFLAGS -O3 -ffat-lto-objects -flto=4 "
export FCFLAGS="$CFLAGS -O3 -ffat-lto-objects -flto=4 "
export FFLAGS="$CFLAGS -O3 -ffat-lto-objects -flto=4 "
export CXXFLAGS="$CXXFLAGS -O3 -ffat-lto-objects -flto=4 "
make  %{?_smp_mflags}

pushd ../buildavx2
export CFLAGS="$CFLAGS -m64 -march=haswell"
export CXXFLAGS="$CXXFLAGS -m64 -march=haswell"
export LDFLAGS="$LDFLAGS -m64 -march=haswell"
make  %{?_smp_mflags}
popd
pushd ../buildavx512
export CFLAGS="$CFLAGS -m64 -march=skylake-avx512 -mprefer-vector-width=512"
export CXXFLAGS="$CXXFLAGS -m64 -march=skylake-avx512 -mprefer-vector-width=512"
export LDFLAGS="$LDFLAGS -m64 -march=skylake-avx512"
make  %{?_smp_mflags}
popd
</pre>

Note that the FCFLAGS/FFLAGS are not set for avx2/avx512 builds, hence any Fortran code is compiled  with "-march=westmere" (inherited from the base build). I don't assume this is by design. AFAIK there is a lot of number crunching Fortran code that would benefit greatly from AVX2 and/or AVX512 instructions.

phmccarty:
This issue was resolved with #598.

Issue #538, opened 2019-11-29
FCFLAGS warnings


jurobystricky:
Autospec assumes Fortran flags are the same as CFLAGS, as can be seen in spec files::

<pre>
export FCFLAGS="$CFLAGS -O3 -ffat-lto-objects -flto=4 "
export FFLAGS="$CFLAGS -O3 -ffat-lto-objects -flto=4 "
</pre>

The assumption is (no longer?) valid, as can be witnessed by the warnings at build time:

<pre>
f951: Warning: command line option ‘-Wformat=1’ is valid for C/C++/ObjC/ObjC++ but not for Fortran
f951: Warning: command line option ‘-Wformat-security’ is valid for C/C++/ObjC/ObjC++ but not for Fortran
</pre>

phmccarty:
This has always been the assumption.

jurobystricky:
Apparently no longer valid assumption. Although the warnings are harmless, the problem is, some builds interpret any warnings as errors.

phmccarty:
This issue was resolved with #598.

PR #598, opened 2020-04-02
specfiles.py: FFLAGS/FCFLAGS fixes

jurobystricky:
Fortran developers may be using either FCFLAGS, FFLAGS or both.
Common usage is:

FCFLAGS : Fortran 90 compiler flags
FFLAGS  : Fortran 77 compiler flags

If both of them are initialized, they must be identical.

This patch addresses two issues:

1. https://github.com/clearlinux/autospec/issues/538
   The assumption CFLAGS and FFLAGS are identical is no longer valid.
   Using CFLAGS for FFLAGS will result in compiler warnings such as:
   f951: Warning: command line option -Wformat=1 is valid for C/C++/ObjC/ObjC++ but not for Fortran
   f951: Warning: command line option -Wformat-security is valid for C/C++/ObjC/ObjC++ but not for Fortran

   Not only are the warnings a nuisance, but if warnings are treated as
   errors, this can break builds. The remedy is to initilize FFLAGS using
   the environment $FFLAGS instead of the environment $CFLAGS.

2. https://github.com/clearlinux/autospec/issues/580
   AVX2/AVX512 instructions not generated for Fortran
   When selecting builds with AVX2/AVX512 optimizations, there are no
   FFLAGS or FCFLAGS specified, hence any resulting Fortran code will
   not generate any AVX instructions. Fortran is mostly used for
   intensive mathematical/numerical/scientific code and as such would
   very likely benefit from AVX instructions a lot.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -576,6 +576,13 @@ class Specfile(object):
                               "-Wl,-z,max-page-size=0x1000 "
                               '-march=westmere -mtune=haswell"\n')
             self._write_strip("export CXXFLAGS=$CFLAGS\n")
+            self._write_strip('export FFLAGS="-O2 -g -Wp,-D_FORTIFY_SOURCE=2 '
+                              "-fexceptions -fstack-protector "
+                              "--param=ssp-buffer-size=32 "
+                              "-Wno-error "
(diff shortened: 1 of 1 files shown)


bryteise:
Nice improvement, thanks!

PR #600, opened 2020-04-06
RFC: package installed tests to foo-tests

rossburton:
Many GNOME packages install their test suite for integration testing[1] but these test suites end up in foo and foo-libexec, thus part of every user's install wasting space.

We could just disable the installed test suite but the tests are actually useful: the builder is too limited an environment for many test suites to actually pass, so running them on a full image is preferable.

This branch does some cleanup to how extras files are parsed, and packages the test suite into foo-tests.  This new package isn't in any bundles, so the files will be removed from images.

[1] https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -391,6 +391,12 @@ dev_extras
   functionality to place files used only for development against this
   software that Autospec does not automatically detect.
 
+tests_extras
+  Same as "extras" above, but instead of the files being placed in an
+  ``-extras`` subpackage, they will be placed in the ``-tests`` one. Use this
+  functionality to place files used only for testing against this
+  software that Autospec does not automatically detect.
+
 ${custom}_extras
   Same as "extras" above, but instead of the files being placed in an
   ``-extras`` subpackage, they will be placed in the ``extras-${custom}``
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -753,30 +753,28 @@ class Config(object):
             print("%%exclude for: %s." % exclude)
         filemanager.excludes += content
 
-        content = self.read_conf_file(os.path.join(path, "extras"))
-        for extra in content:
-            print("extras for  : %s." % extra)
-        filemanager.extras += content
-
         for fname in os.listdir(path):
-            if not re.search('.+_extras$', fname) or fname == "dev_extras":
+            if re.search(r'.+_extras$', fname):
+                # Prefix all but blessed names with extras-
+                name = fname[:-len("_extras")]
+                if name not in ('dev'):
+                    name = f'extras-{name}'
+            elif fname == 'extras':
+                name = 'extras'
+            else:
                 continue
+
             content = {}
             content['files'] = self.read_conf_file(os.path.join(path, fname))
             if not content:
                 print_warning(f"Error reading custom extras file: {fname}")
                 continue
+
             req_file = os.path.join(path, f'{fname}_requires')
             if os.path.isfile(req_file):
                 content['requires'] = self.read_conf_file(req_file)
-            name = fname[:-len("_extras")]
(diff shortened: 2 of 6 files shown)


bryteise:
Would it be reasonable to have the tests subpackage be only the fixed paths and then use custom extras (and maybe custom extra requires) to get the left over files? Maybe I should look at a few specific packages.

I somewhat want to avoid having a path pick up files for a subpackage and have manual configuration for the same package.

rossburton:
Does it make more sense if viewed from the other direction: tests that can be run installed (e.g. fwupd ships mock pieces so you can do integration testing) should be packaged into `foo-tests`.  These files can be listed in `tests_extras`.  For convenience, the GNOME tradition of `/usr/*/installed-tests` is packaged automatically.

Concrete examples include gdk-pixbuf (ships a test suite in `installed-tests`, entirely packaged by autospec) and mutter (does that, but also ships test data in another directory so needs `tests_extras`).

autospec/config.py
```diff
@@ -753,30 +753,29 @@ def parse_config_files(self, path, bump, filemanager, version, requirements):
...
                 content['requires'] = self.read_conf_file(req_file)
-            name = fname[:-len("_extras")]
-            print(f"extras-{name} for {content['files']}")
-            filemanager.custom_extras["extras-" + f"{name}"] = content
 
-        content = self.read_conf_file(os.path.join(path, "dev_extras"))
-        for extra in content:
-            print("dev for     : %s." % extra)
-        filemanager.dev_extras += content
+            print(f"{name} for {content['files']}")
```
bryteise, line 777: I think this might no longer be a valuable thing to show for each custom extra file. Probably can drop printing these at this point.

tests/test_files.py
```diff
@@ -156,25 +156,13 @@ def test_push_file_autostart(self):
...
-
-
     def test_push_file_custom_extras(self):
         """
         Test push_file to a custom extras package
         """
         self.fm.file_is_locale = MagicMock(return_value=False)
         self.fm.push_package_file = MagicMock()
-        self.fm.custom_extras = {'test-extras': {'files': ["test"]}}
+        self.fm.file_maps = {'test-extras': {'files': ["test"]}}
```
bryteise, line 165: Seems like an easily confused custom name since test-extras is very close to tests-extras which is special given the other parsing in this PR.

bryteise: (I know this wasn't your choice of name just maybe we should change it)

README.rst
```diff
@@ -391,6 +391,12 @@ dev_extras
   functionality to place files used only for development against this
   software that Autospec does not automatically detect.
 
+test_extras
```
bryteise, line 394: Should be tests and simimlarly -tests below given the rest of the PR.

bryteise:
Okay, I think this is good. It doesn't play nice with the custom requires which it probably should but I have another change I need to do for that anyway so not really something to change here.

bryteise:
I also think that the tests probably should depend on the dev subpackage to be useful? Maybe bin and lib too as some packages may not have dev.

autospec/config.py
rossburton: All of the debugging information or just this one?

rossburton:
-tests don't need to depend on -dev as they're already compiled.  I've added an extra dependency so that foo-tests depends on foo (like -dev does).

autospec/config.py
bryteise: Probably anything config related the user sets themselves but that's another change. For this any of the extras prints can safely be silenced.

bryteise: My goal is to avoid a huge amount of spamming when extras are configured at least.

PR #602, opened 2020-04-06
specfiles: Allow archives in R pattern

anselmolsm:
This is mainly to support extra tarball for R-RSQLite, to bump the
bundled sqlite version using upstream tarball instead of a local (huge) patch.


bryteise:
Ugh this code (not your change) needs to be redone so badly. The nested logic is not something that makes much sense at this point.

PR #604, opened 2020-04-11
Make the target directory part of the Config class

bryteise:


autospec/pkg_integrity.py
```diff
@@ -940,7 +940,7 @@ def load_specfile(specfile):
 
 def main(args):
     """Verify tar content with signature."""
-    from_disk(args.url, args.tar, args.sig, config.Config())
+    from_disk(args.url, args.tar, args.sig, config.Config(""))
```
phmccarty, line 943: Hmm, I think I would prefer removing the `main` function entirely at this point, since all references to `config.download_path` in the rest of the file will yield unexpected results.

bryteise: It was a handy idea but we really would need to improve configuration handling to bring it back so dropping it seems like the right move at this point.

tests/test_pkg_integrity.py
```diff
@@ -47,30 +47,30 @@ def mock_download_do_curl(url, dst=None):
...
+    def _mock_rewrite(self):
+        pass
 
     def test_check_matching_sign_url(self):
-        conf = config.Config()
-        conf.rewrite_config_opts = self._mock_rewrite
-        conf.config_opts['verify_required'] = False
         with tempfile.TemporaryDirectory() as tmpd:
+            conf = config.Config(tmpd)
+            conf.rewrite_config_opts = self._mock_rewrite
```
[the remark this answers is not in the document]
bryteise: Ah great idea, thanks.

tests/test_pkg_integrity.py
```diff
@@ -4,7 +4,8 @@
 import shutil
 import tempfile
 import unittest
-from unittest.mock import patch
+from unittest import mock
+from mock import patch
```
phmccarty, line 8: Shouldn't we import from `unittest.mock`? `mock` is a backport package.

bryteise: Ah

bryteise: Didn't realize I was shadowing lol

bryteise: This passed on my system without issue and so I didn't think about it much.

PR #605, opened 2020-04-12
Extend fail pattern for `make`

phmccarty:
This covers the "ENOENT" case. Also update the unittests to verify that
the updated pattern works.


bryteise:
Oh nice.

PR #615, opened 2020-04-17
config: only load profile_payload if PGO is enabled

### LINKED ISSUE #548: profile_payload gets read even if pgo is false
As reported by @bwarden  the behavior of the pgo flag in options.conf can be inconsistent. It looks like if a `profile_payload` file exists in the directory it gets read and used, even if `pgo = false` is set in options.conf 

This scenario could occur if you tried autospec a package with pgo but later decided they did not want to use it or encountered bugs.

A more consistent behavior would be to have the `profile_payload` file only read if `pgo = true`. A warning could be displayed to the user if `pgo = false` but the `profile_payload` file exists or vice versa. 

Sources:
https://github.com/clearlinux/autospec/blob/196235ca223da095d58a604a93dfdde336106ad1/autospec/config.py#L978

https://github.com/clearlinux/autospec/blob/cd2fcdfc595eaba2974fefee0adc7fc6754b82f5/autospec/specfiles.py#L623

rossburton:
Closes #548

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -920,7 +920,8 @@ class Config(object):
         self.install_append = self.read_script_file(os.path.join(self.download_path, "install_append"))
         self.service_restart = self.read_conf_file(os.path.join(self.download_path, "service_restart"))
 
-        self.profile_payload = self.read_script_file(os.path.join(self.download_path, "profile_payload"))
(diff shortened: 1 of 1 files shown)


bryteise:
Yay.

PR #619, opened 2020-04-22
Rework for color-coded progress messages

phmccarty:
- Remove the progress message functions from `pkg_integrity`, and extend/refactor the ones from `util` to support the needs of `pkg_integrity`.
- Avoid printing terminal escape sequences for color when stdout is not a tty.

### DIFF
--- a/autospec/util.py
+++ b/autospec/util.py
@@ -22,6 +22,7 @@ import os
 import re
 import shlex
 import subprocess
+import sys
 
 dictionary_filename = os.path.dirname(__file__) + "/translate.dic"
 dictionary = [line.strip() for line in open(dictionary_filename, 'r')]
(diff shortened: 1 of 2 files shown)


autospec/util.py
```diff
@@ -89,19 +89,52 @@ def get_sha1sum(filename):
...
+        elif color == 'blue':
+            params = '34;1'
+        prefix = f'\033[{params}m{level}\033[0m'
+    print(f'[{prefix}] {message}')
 
 
-def print_infile(message):
-    """Print colored INIFILE content."""
-    print("[\033[1;34mINFILE\033[0m] {}".format(message))
+def print_error(message, level='ERROR'):
```
bryteise, line 114: It seems odd to me these functions would take a level parameter.

Why not just set level in the function body (or not at all and pass it as a literal like color)?

phmccarty: It is a leftover from a previous iteration of my patch. I will remove the keyword parameters.

phmccarty: Done

bryteise:
LGTM

PR #620, opened 2020-04-23
Add support for custom subpackage requires

bryteise:
Due to more systems being added for automated runtime dependency
injection for subpackages, we need a way for manual overrides of the
default detected dependencies.

This change provides support for specifying which subpackage runtime
dependencies should be added or removed and consolidates some of the
custom configuration file detection along with it.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -555,9 +555,6 @@ class Config(object):
         """Process extras type subpackages configuration."""
         content = {}
         content['files'] = self.read_conf_file(os.path.join(self.download_path, fname))
-        if not content:
-            print_warning(f"Error reading custom extras file: {fname}")
-            return
         req_file = os.path.join(self.download_path, f'{fname}_requires')
         if os.path.isfile(req_file):
             content['requires'] = self.read_conf_file(req_file)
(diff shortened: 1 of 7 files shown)


phmccarty:
@bryteise I just tested the code with flake8-3.8.0a2 (and pycodestyle-2.6.0a1), and no issues are identified with your changes. There are 4 new issues raised, but I'll address those in a different PR.

Proceeding with review...

autospec/config.py
```diff
@@ -551,6 +551,28 @@ def read_conf_file(self, path, track=True):
         lines = self.read_file(path, track=track)
         return [l.strip() for l in lines if not l.strip().startswith("#") and l.split()]
 
+    def process_extras_file(self, fname, name, filemanager):
+        """Process extras type subpackages configuration."""
+        content = {}
+        content['files'] = self.read_conf_file(os.path.join(self.download_path, fname))
+        if not content:
```
phmccarty, line 558: I see that this code lived elsewhere before, but will this condition ever be true? `content['files']` will be populated whether or not there is an error (in which case it is `[]`).

bryteise: Yea, this condition should be fixed. I'll capture that in another PR.

PR #626, opened 2020-04-26
Requirements: Properly initialize the class

### LINKED ISSUE #625: AttributeError raised for pypi_provides
With recent changes to the Requirements class in support of adding/banning deps at the subpackage level, I am running into an exception:

```
Traceback (most recent call last):
  File "../../projects/autospec/autospec/autospec.py", line 355, in <module>
    main()
  File "../../projects/autospec/autospec/autospec.py", line 215, in main
    package(args, url, name, archives, workingdir, infile_dict)
  File "../../projects/autospec/autospec/autospec.py", line 302, in package
    specfile.write_spec()
  File "/home/pmccarty/clearlinux/projects/autospec/autospec/specfiles.py", line 90, in write_spec
    self.write_files_header()
  File "/home/pmccarty/clearlinux/projects/autospec/autospec/specfiles.py", line 290, in write_files_header
    if self.requirements.pypi_provides:
AttributeError: 'Requirements' object has no attribute 'pypi_provides'
```

bryteise:
The pypi_provides memeber wasn't being set on initialization causing
an error when processing the specfile in some cases.

Fixes #625.

### DIFF
--- a/autospec/buildreq.py
+++ b/autospec/buildreq.py
@@ -284,6 +284,7 @@ class Requirements(object):
         self.extra_cmake_openmpi = set()
         self.verbose = False
         self.cargo_bin = False
+        self.pypi_provides = None
         self.banned_buildreqs = set(["llvm-devel",
                                      "gcj",
                                      "pkgconfig(dnl)",


bryteise:
I don't have a test for this one because it seems weird to test class initialization routines.

autospec/buildreq.py
```diff
@@ -284,6 +284,7 @@ def __init__(self, url):
         self.extra_cmake_openmpi = set()
         self.verbose = False
         self.cargo_bin = False
+        self.pypi_provides = None
```
phmccarty, line 287: Hmm, should we store this metadata in `provides["pypi"]` instead? I expect that would require more rework though...

phmccarty: On second look, I don't see the `"pypi"` key used for either `provides` or `requires`... Then to rephrase my question: Should we store the pypi provides metadata in `provides["python3"]`?

bryteise: We should once we rework provides like I did the requires X(.

phmccarty: Okay, I opened #627 to track that work.

PR #636, opened 2020-05-13
Ignore "unknown xref" warnings for R packages

phmccarty:
These warning messages appear whenever an R package's tests cannot
resolve documentation cross-references to other packages.

Although the warnings may create usability problems with a package's
documentation, any unresolvable cross-references are for *optional*
or *undeclared* dependencies, and thus do not impact the overall
functionality of that package.

I consider these unresolvable cross-references to be a minor drawback,
and there is greater benefit to ignoring them than acting on them: (a)
build rounds for R packages will be reduced in general, especially
improving build times for packages with multiple warnings raised, and
(b) the dependency tree becomes smaller, simplifying and improving the
time to rebuild all R packages in dependency order.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -293,7 +293,6 @@ class Config(object):
             (r"Target '[a-zA-Z0-9\-]' can't be generated as '(.*)' could not be found", 0, None),
             (r"Unable to `import (.*)`", 0, None),
             (r"Unable to find '(.*)'", 0, None),
-            (r"Unknown packages? ['‘]([a-zA-Z0-9\-]*)['’].* in Rd xrefs", 0, 'R'),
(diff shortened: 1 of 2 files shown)


autospec/config.py
```diff
@@ -293,7 +293,6 @@ def __init__(self, download_path):
             (r"Target '[a-zA-Z0-9\-]' can't be generated as '(.*)' could not be found", 0, None),
             (r"Unable to `import (.*)`", 0, None),
             (r"Unable to find '(.*)'", 0, None),
-            (r"Unknown packages? ['‘]([a-zA-Z0-9\-]*)['’].* in Rd xrefs", 0, 'R'),
```
bwarden, line 296: smartquotes ftw

bwarden: Maybe a + instead of * here though, just to be slightly more careful.

phmccarty: This is a line I removed... But yes, I agree, if we were keeping the pattern, a `+` would be better.

bwarden: ahhh, quarantine brain. sorry.

PR #639, opened 2020-06-07
Add nomissingbuildids, noautoreq, noautoprov flags

insilications:
New useful controlling flags for options.conf that helps to build certain packages

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -597,6 +597,17 @@ compat
 
 nodebug
   If this option is set, ``debuginfo`` is not created for this package.
+  
+nomissingbuildids
+  If this option is set, missing build ids will not terminate the build.
+  
+noautoreq
+  If this option is set, ``AutoReq: no`` will be set for this package and 
+  automatic requeriments processing will be disabled
+  
+noautoprov
+  If this option is set, ``AutoProv: no`` will be set for this package and
+  automatic provides processing will be disabled
 
 Name and version resolution
 ===========================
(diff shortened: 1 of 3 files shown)


fenrus75:
I'm not immediately a fan...
these flags often end up with broken packages, and are likely an indication of something much bigger going on where the package is really not good for autospec.

maybe examples of actual things can convince me, but I'm not a fan on making broken packages....

insilications:
I created an autospec'd package for the latest installed CUDA version in order to populate my local repo and build some machine learning packages with mock. rpmbuild fails for CUDA (missing build ids and creates unnecessary dependency requirements when it looks into some of the shared libraries provided by CUDA, such as Qt 5.9). You are probably right about enabling the build of broken packages. I thought this could help people in some other use cases.

fenrus75:
I'm not too worried about the build-id thing, we even had that by default for some time
(it just means you do not get any debug info but well.. so be it)

the requirements is what worries me..... that's directly leading to broken packages.

if you have a false requirement that is truely false, a workaround is to make a dummy package in your repo that just Provides: that one thing.

the autoreqprov is for ALL dependencies, too big a hammer normally compared to a very narrow 1 or 2 unfortunate ones.

fenrus75:
I can imagine a "fakeprovides" file in autospec where any truely false requirement can just be Provided: directly.... that's much more surgical than turning off all requires

insilications:
> if you have a false requirement that is truely false, a workaround is to make a dummy package in your repo that just Provides: that one thing.

> I can imagine a "fakeprovides" file in autospec where any truely false requirement can just be Provided: directly.... that's much more surgical than turning off all requires

Yes, makes sense. This approach is much more safe and precise.

PR #640, opened 2020-06-09
Add new cmake module mappings

phmccarty:


phmccarty:
The file sort was botched... I will fix that shortly.

Issue #646, opened 2020-08-31
multiple problems regarding BitBake file parsing


lebensterben:
## Operator `??=` doesn't seem to be escaped correctly
Should that be `\\?\\?\\=`?
https://github.com/clearlinux/autospec/blob/b110b315e2bec7fb9629ca176b5bb59a05b70f2f/autospec/infile_bb_parser.py#L124

## Parsing order of `.inc` and `.bb` files are incorrect.
BitBake files (`.bb`, `.bbclass`, or `.inc`) uses [`include` directive] or [`require` directive] to include the contents of a `.inc` file, at the point of the directive. Suppose we have a `foo.inc`:
```bb
A = "1"
```
And if there's a `bar.bb` s.t.:
```bb
A = "2"
include foo.inc
```
Then `A` should evaluate to `"2"`. But the current implementation parses `.inc` files first, that would make `A` equal to `"1"`.

## File structures is incorrect for `.inc` files ( and `.bbclass` files )

[The documents says][file structure]
> In order for include and class files to be found by BitBake, they need to be located in a "classes" subdirectory that can be found in BBPATH.

For a correct `.bb` file which includes or requires a `.inc` or `.bbclass` file, it would put them in the correct directory. But the current implementation doesn't reads into sub-directories.

## Missing features
First, `_prepend` and `_append` is only implemented for for shell functions, but not variables.

Then, the more general `override` syntax, including [`_remove`] and [conditionals] are not implemented. These are quite fundamental.

Third, it's already stated in comments that Python functions are not scraped, which includes BitBake Style Python function, Python funcition, and anonymous Python function. It's better to put this information in manpage or documentation.

[`include` directive]: https://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#include-directive
[`require` directive]: https://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#require-inclusion
[file structure]: https://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#locating-include-and-class-files
[`_remove`]: https://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#removing-override-style-syntax
[conditionals]: https://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#conditional-syntax-overrides

lebensterben:
[...]

Issue #645, opened 2020-08-29
question: how infile handler works?


lebensterben:
It's documented that `autospec` could accept a directory of URLs as argument to `--infile`, which is handled by `infile_reader` function:

https://github.com/clearlinux/autospec/blob/b110b315e2bec7fb9629ca176b5bb59a05b70f2f/autospec/infile_handler.py#L130-L140

But I cannot understand how this could work.

`infile_reader` calls `file_handler`, which in turn calls `check_url_content`:

https://github.com/clearlinux/autospec/blob/b110b315e2bec7fb9629ca176b5bb59a05b70f2f/autospec/infile_handler.py#L146-L147

https://github.com/clearlinux/autospec/blob/b110b315e2bec7fb9629ca176b5bb59a05b70f2f/autospec/infile_handler.py#L112-L114

`check_url_content` sends a HEAD request to the `url`, which is the file name of a infile in a given directory.

https://github.com/clearlinux/autospec/blob/b110b315e2bec7fb9629ca176b5bb59a05b70f2f/autospec/infile_handler.py#L64

The problem is, for a valid HTTP or HTTPS URL, it must contain the scheme `http://` or `https://`. And `/` is not a reserved character on most Linux file systems. That is, `/` cannot appear in the file name, or in the `url` passed to `check_url_content`.

The implication is,

- Suppose I do have a directory with infiles whose file names are some kind of 'URL's. They won't be valid HTTP(s) URL. And thus they won't be treated correctly. 

I think maybe this python module is not completely/correctly implemented. Correct me if I'm wrong.

---

P.S. From the commit message https://github.com/clearlinux/autospec/commit/aef3399d1058daea96af38fd86742866ece7f512, it reads that
> The --infile argument now allows a url, file, or directory of files to be passed as the input.

So probably, it was never intended to deal with a directory of URLs.

phmccarty:
@lebensterben This is a feature we started to implement for autospec but ultimately, we never used it. So, if you see any problems with the docs or code relevant to the feature, that is why.

phmccarty:
Resolved with #647

PR #647, opened 2020-09-01
Removed `infile` features

lebensterben:
There are multiple issues regarding `infile` support.

Notably,
- It doesn't support some basic syntax of `BitBake`.
- It doesn't fully comply to `BitBake`'s conventions.
- The differences from `BitBake` is not well-documented.

Since this is never used, I propose to remove this functionality entirely.

https://github.com/clearlinux/autospec/issues/645
https://github.com/clearlinux/autospec/issues/646


phmccarty:
@lebensterben I am okay with removing these features, but I would prefer not to see all of the code style changes interleaved in the same commit... Did your editor make the style changes automatically?

If you would like to make some changes to code style, make a separate commit (with rationale in the commit description if necessary), and make sure that the changes apply to the entire code base. Or you can omit the style changes from this PR if you prefer.

lebensterben:
Yeah. emacs applied reformatted some files according and reordered imports.

lebensterben:
It's done now.
BTW, would you add a `.editorconfig` or `.flake8` file to the root, so that modern editors can use your per-project style settings?

phmccarty:
The flake8 config already enforced by Github Actions. You can run the checks locally by running `make check`. The behavior is controlled via the toplevel `setup.cfg`.

I like the idea of adding a `.editorconfig`, since it can cover basic style issues. Opened an issue for that: #648

phmccarty:
@lebensterben It looks like all test files under `tests/testfiles/bb/` can be removed as well? Can you check that?

Otherwise, changes look good to me.

lebensterben:
@phmccarty I removed them (`tests/testfiles/bb`) in the new commit.

Issue #656, opened 2021-04-30
Stop using `pip search`


phmccarty:
The `pip search` command is slated to disappear eventually, and it's been nonfunctional for a few months due to the pypi xmlrpc server being disabled.

We should stop using that command and find a functional replacement.

(I'm labeling this as a bug because we rely on `pip search` as part of the addition of `pypi(...)` virtual provides/requires.)

phmccarty:
Fixed with #657

PR #657, opened 2021-04-30
Use pypi json API instead of `pip search`

phmccarty:
`pip search` is currently non-functional. Instead we can query the pypi json
API for package metadata.

If a pypi package with the specified name does not exist, the API returns HTTP
404. Otherwise, a json reponse is returned. Redirects may occur if the
specified name maps to a different pypi package name. We don't have to bother
handling that, because redirects will be followed automatically by
`download.do_curl()`.

### DIFF
--- a/autospec/pypidata.py
+++ b/autospec/pypidata.py
@@ -6,6 +6,8 @@ import subprocess
 import sys
 import tempfile
 
+import download
+
 
 def pip_env():
     """Generate a copy of os.environ appropriate for pip."""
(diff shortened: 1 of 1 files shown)


phmccarty:
This appears to work well after testing a few packages. I want to test a bit more before considering this approach though.

phmccarty:
Fixes #656

phmccarty:
Okay, testing looks good :-)

fenrus75 (approved):

PR #666, opened 2021-07-30
golang: support both Makefile and makefile

mdhorn:
We should check for both 'makefile' and 'Makefile' for go-lang projects.


phmccarty (approved):
+1

In the future, it would be nice to implement a "supported makefile" regular expression and re-use that throughout the codebase, but I'm okay with adding the additional check for now.

PR #671, opened 2021-09-28
add a server flag in options.conf that triggers outright avx2 use:

fenrus75:


phmccarty:
PR #672 supercedes this one, so I'll close it.

PR #673, opened 2021-09-29
Don't handle install_requires parsing when already in multiline

bryteise:
When parsing setup.py files, if already in multiline and install_requires
is used as a variable, don't try and treat the line as if it was an
install_requires assignment.


fenrus75 (approved):

phmccarty (commented):
Already merged, but +1 :-)

PR #674, opened 2021-10-04
Print unit test errors for meson pattern

phmccarty:
The `meson test` subcommand has a `--print-errorlogs` flag that we can
use to print details about any failed unit tests. Thus any errors will
be more easily discoverable by inspection of the mock build.log.


bryteise:
Nice!

PR #688, opened 2021-10-19
Add pypi overrides

bryteise:
See https://github.com/clearlinux/clr-avx-tools/commit/013a989f5ad273adbc00e34eb18b87984b04ba78 for details about the new script autospec started to use for doing the dependency modifications.

Also adds some minor autospec cleanups I found along the way.


bryteise:
@phmccarty Yep, I poked the right UI buttons to enable.

PR #707, opened 2022-01-05
Avoid rebuilding the package after only file adds

bryteise:
Try to avoid doing a full rebuild when the build.log only has new
files in it.

This has a two downsides. The source buildroot needs to be distinct
from the binary buildroot. The built rpms are no longer installable as
rpmbuild flags them.

The upside is packages that take forever to build or don't have
build times improved by ccache will not need a duplicate round for
just adding the files.

The rpms can still be generated in an installable fashion with 'make
build' and the autospec generated rpms can be installed by force
installing if need be.

### DIFF
--- a/autospec/autospec.py
+++ b/autospec/autospec.py
@@ -285,7 +285,7 @@ def package(args, url, name, archives, workingdir):
             # directories added to the blacklist, need to re-run
             package.must_restart += 1
 
-        if package.round > 20 or package.must_restart == 0:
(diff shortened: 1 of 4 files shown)


bryteise:
I've been using this for a few weeks now and hopefully have run into most of the edge cases but there may be a few warts left over. It also may just not be worthwhile to make this optimization.

insilications:
What about patching rpmbuild to avoid the short-circuit flags? I use it here

bryteise:
@insilications Yea, that is something we can do and I'd be happy with that change.

PR #711, opened 2022-01-21
Fix test processing

bryteise:
If a file addition only failure happens and a quick rebuilt is done,
the build log will be missing test results. Check the previous build
log in that case.

### DIFF
--- a/tests/test_check.py
+++ b/tests/test_check.py
@@ -42,7 +42,7 @@ class TestTest(unittest.TestCase):
         m_open = mock_open()
         open_name = 'util.open'
         with patch(open_name, m_open, create=True):
-            check.check_regression('pkgdir', False)
+            check.check_regression('pkgdir', False, -1)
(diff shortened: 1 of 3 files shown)


tests/test_check.py
```diff
@@ -42,7 +42,7 @@ def mock_parse_log(log):
         m_open = mock_open()
         open_name = 'util.open'
         with patch(open_name, m_open, create=True):
-            check.check_regression('pkgdir', False)
+            check.check_regression('pkgdir', False, -1)
```
phmccarty, line 45: I would prefer for the new argument to be `1`, since the `-` is treated as a hyphen in `check_regression`, and it looks like a "negative 1" here at the call site.

phmccarty: ignore.. I misread the code.

phmccarty (approved):
+1

PR #712, opened 2022-01-21
Setuid opt skip

bryteise:


phmccarty (approved):

phmccarty:
@bryteise Do you recall which packages needed fixups that prompted the changes in #684 ?

bryteise:
Hrm no, we might need another target at some point but perhaps the issue from #684 was resolved in another way.

fenrus75:
we do not have that many setuid at least

also the ordering wrt install_append worries me; now we cannot fix up things before the files are copied away

bryteise:
We can fixup after copy but it is harder to do potentially though if we are build/installing multiple times we likely should make the fix pre-install rather than append but that can be easier said than done. I'll undo.

phmccarty:
This change to mariadb might have been relevant: https://github.com/clearlinux-pkgs/mariadb/commit/1d18c2be715578b5570118c6cd14708f9b62f6cd

phmccarty:
And also: https://github.com/clearlinux-pkgs/nettle/commit/d6e160e710356cf2dca2365528573878a423b0a1

PR #721, opened 2022-02-07
Don't use beignet for opencl

bryteise:
beignet is not used for the latest opencl hardware so stop auto adding
it as a dependency.


phmccarty (commented):
+1

PR #724, opened 2022-02-15
commitmessage: support more exact version prefixes

phmccarty:
It's common to see git tag names with a `<packagename>-` prefix for the
version string, also perhaps appearing before a `v` or `V`.


bryteise:
Nice

PR #728, opened 2022-03-02
Create a new object for multi_version

### LINKED ISSUE #727: The `versions` output is incorrect when version cannot be parsed
For example, create a new package sourced from https://github.com/nipy/nibabel/archive/master.zip, and autospec will create a `versions` file with the following content:

```
1       https://github.com/nipy/nibabel/archive/master.zip
```

(`1` is the default version set if the version cannot be parsed from the URL...)

Instead, the file should only contain the version:

```
1
```

bryteise:
When not handling a multiple version package, there was a case where if
the version wasn't set the multi_version map would be the
config.versions map and that's not intended to outside of multiple
version use case.

So to fix this, make a new map when setting a default version for the
multi_version map and keep config.versions as its own map.

Fixes #727 

### DIFF
--- a/autospec/tarball.py
+++ b/autospec/tarball.py
@@ -243,8 +243,9 @@ class Content(object):
         elif not self.multi_version:
             # Fall back to ensure a version is always set
             # (otherwise the last known version will be used)
-            self.multi_version["1"] = ""
+            self.multi_version = {"1": ""}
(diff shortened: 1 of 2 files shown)


phmccarty (approved):
+1, thanks!

PR #731, opened 2022-03-17
Don't exit on missing dependencies

### LINKED ISSUE #729: buildreq cache not caching fail pattern dependencies
Attempting to update gst-plugins-bad from 1.20.0 to 1.20.1:

- autospec detects several build dependencies over the course of build rounds 1-3
- on build round 4, two unknown build dependencies are detected
- autospec exits with an error

The spec file looks correct, but `buildreq_cache` is not created... It should contain the build dependencies detected (via fail patterns) in build rounds 1-4. This was a feature added in https://github.com/clearlinux/autospec/pull/335, and it was still working in the not-too-distant past...

bryteise:
Previous change made missing build dependendies fatal, and while
correct, causing the buildreq_cache file to not be updated when
missing dependencies cause the build to exit.

Move back to previous obvserved behavior but change the logic to
return if the buildroot log is not successful.

Fixes #729 

### DIFF
--- a/autospec/build.py
+++ b/autospec/build.py
@@ -300,7 +300,7 @@ class Build(object):
             exit(1)
 
         if not self.parse_buildroot_log(config.download_path + "/results/root.log", ret):
-            exit(1)
+            return
 
         self.parse_build_results(config.download_path + "/results/build.log", ret, filemanager, config, requirements, content)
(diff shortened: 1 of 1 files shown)


phmccarty (approved):
LGTM

Thanks!

PR #734, opened 2022-04-04
tests: for every package URL, test all content states

phmccarty:
The generated packageurl name/version tests were only testing one of four content states, leading to a gap in test coverage, since some names or versions might be mis-parsed in a state that was not tested.

Fix the issue by testing all content states for every package URL in the tests/packageurls file. And ensure that test failure output remains actionable by using a subTest() context manager and setting the `state` argument, which will print the value of `state` in the test failure output header.

### DIFF
--- a/tests/test_tarball.py
+++ b/tests/test_tarball.py
@@ -130,31 +130,40 @@ def source_test_generator(url, destination, path, content, src_type, prefix, sub
     return generator
 
 
-def name_and_version_test_generator(url, name, version, state):
+def name_and_version_test_generator(url, name, version):
     """Create test for tarball.name_and_version method."""
     def generator(self):
         """Test template."""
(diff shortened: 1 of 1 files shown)


fenrus75 (approved):

phmccarty:
I discovered this issue when exploring a bug fix for the version parsing code... adding 2 new packageurls (before patching the code to fix the bug) with the "expected versions", only one test failed instead of two.

Then I noticed how these tests are generated, cycling through the four different states, and I figured it would be better just to test all states for every packageurl. The test that mistakenly passed had an explicit version arg set instead of being autodetected...

If this PR looks okay, I will follow up with that bug fix.

bryteise:
Ah much more complete, looks good.

PR #737, opened 2022-04-14
Update fail pattern for pypi-docutils

phmccarty:


bryteise:
Oops missed that one, thanks!

PR #738, opened 2022-04-16
config: improve handling of backport patches

phmccarty:
If a patch applies changes to more than one file, multiple `Skipping patch` messages may appear in the log. For that case, multiple calls to `remove_backport_patch()` will occur. Support that case by returning early if the patch has already been removed from `self.patches`.

Also, add a progress message to indicate a backport patch has been removed, otherwise no progress message is likely to appear between build rounds.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -29,7 +29,7 @@ from collections import OrderedDict
 
 import check
 import license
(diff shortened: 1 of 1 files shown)


bryteise:
Nice!

PR #742, opened 2022-05-04
Refactor pytest test summary parsing; update tests

phmccarty:
Instead of using separate regular expressions to capture the multiple
forms pytest test summaries might take, use one generic regular
expression to match the summary line, and then process comma-separated
components of the line with more specific matches.

Three test cases needed to be updated to accommodate this refactor:

- Two tests omitted the leading '=' characters from the line to match,
  even though their summary lines clearly originated from pytest. Fix by
  adding some leading '=' characters for those summary lines.

- One test mistakenly counted an "xpass" as an "xfail". Autospec maps
  "xpass" results to "pass", so update the test accordingly.

### DIFF
--- a/autospec/count.py
+++ b/autospec/count.py
@@ -254,120 +254,29 @@ def parse_log(log, pkgname=''):
             counted_pass += 1
             continue
 
-        # cryptography
-        # ================= 76230 passed, 267 skipped in 140.23 seconds ==================
-        match = re.search(r"== ([0-9]+) passed, ([0-9]+) skipped in ", line)
+        # for packages using pytest...
+        match = re.search(r"== .*[0-9]+ x?(?:failed|passed|skipped|warnings|pytest-warnings|error) in [0-9.]+(?:s| seconds)", line)
         if match and incheck:
-            total_pass += convert_int(match.group(1))
-            total_skip += convert_int(match.group(2))
-            continue
-
-        # ============= 1628 passed, 72 skipped, 4 xfailed in 146.26 seconds =============
-        match = re.search(r"== ([0-9]+) passed, ([0-9]+) skipped, ([0-9]+) xfailed in ", line)
-        if match and incheck:
-            total_pass += convert_int(match.group(1))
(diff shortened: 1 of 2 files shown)


phmccarty:
Needs a bit more work, so I converted this PR to a draft state.

phmccarty:
Okay, I'm happier with the current revision now. Converting back to a normal PR.

bryteise:
Lovely cleanup. I'd be inclined to turn the pytest warnings into fail considering what we do with warning in general but I don't know how much breakage we'd get from that change.

bryteise (approved):

phmccarty:
> Lovely cleanup. I'd be inclined to turn the pytest warnings into fail considering what we do with warning in general but I don't know how much breakage we'd get from that change.

If we count the pytest-warnings towards the fail count, we would just need to update a few of the test cases. I added the FIXME because they have not counted as fails before, but I would be open to changing it.

bryteise:
> If we count the pytest-warnings towards the fail count, we would just need to update a few of the test cases. I added the FIXME because they have not counted as fails before, but I would be open to changing it.

Oops sorry, I was meaning how many packages would "fail" to build because tests now have a failure in them where they did not before.

phmccarty:
> > If we count the pytest-warnings towards the fail count, we would just need to update a few of the test cases. I added the FIXME because they have not counted as fails before, but I would be open to changing it.
> 
> Oops sorry, I was meaning how many packages would "fail" to build because tests now have a failure in them where they did not before.

Ah, I didn't think that any changes to autospec's test stats collection would affect the package builds... IIRC, the exit status from the script run in `%check` was the only factor influencing the success or failure there.

bryteise:
> Ah, I didn't think that any changes to autospec's test stats collection would affect the package builds... IIRC, the exit status from the script run in `%check` was the only factor influencing the success or failure there.

Hrm I don't know why I thought it would, looking at the code check_regression isn't a failure if (and doesn't even really compare) it just will make it obvious from the diff in test counts things changed. Sorry for the noise.

PR #752, opened 2022-06-22
pypidata: print stderr output for any failed subprocesses

phmccarty:
In the event that the virtualenv or pip subprocess commands from this module return an error code, print more information obtained from stderr to help diagnose the issue(s).


djklimes (approved):
More information is good :D

PR #754, opened 2022-06-29
Enable provides configuration overrides

bryteise:
Add support for (*_)provides_{add,ban} configuration files.

This is occasionally useful for cases where a pypi() provide isn't
picked up as the packages isn't a python package but does provide a
python library. It also can be handy when doing some renames to
prevent mass rebuilds.

### DIFF
--- a/tests/test_buildreq.py
+++ b/tests/test_buildreq.py
@@ -80,6 +80,36 @@ class TestBuildreq(unittest.TestCase):
         self.assertFalse(self.reqs.add_requires('testreq', []))
         self.assertNotIn('testreq', self.reqs.requires[None])
 
+    def test_ban_provides(self):
+        """
+        Test ban_provides with prov already present in provides
+        """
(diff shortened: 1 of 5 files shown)


tests/test_buildreq.py
```diff
@@ -80,6 +80,36 @@ def test_add_requires_not_in_buildreqs(self):
...
+
+    def test_ban_provides_subpkg(self):
+        """
+        Test ban_provides on a subpkg with prov already present in provides
+        """
+        self.reqs.provides['subpkg'] = set(['testreq'])
+        self.reqs.ban_provides('testreq', subpkg='subpkg')
+        self.assertNotIn('testreq', self.reqs.provides['subpkg'])
+
+    def test_add_requires(self):
```
phmccarty, line 99:
```suggestion
    def test_add_provides(self):
```

phmccarty (commented):
LGTM aside from a unit test function to rename (see inline suggestion).

PR #758, opened 2022-08-05
pkg_integrity: verify sha256 instead of md5 for pypi packages

phmccarty:
All sdist tarballs available on pypi appear to have both an md5sum and
sha256sum listed, so switch to verify the stronger hash (sha256)
instead.

### DIFF
--- a/autospec/pkg_integrity.py
+++ b/autospec/pkg_integrity.py
@@ -443,12 +443,13 @@ class QtIoVerifier(ShaSumVerifier):
 
 
 # PyPi Verifier
-class PyPiVerifier(MD5Verifier):
-    """Verify MD5 signature for pypi."""
+class PyPiVerifier(ShaSumVerifier):
+    """Verify SHA256 digest for pypi."""
 
     def __init__(self, **kwargs):
-        """Passthrough initialization to MD5."""
-        MD5Verifier.__init__(self, **kwargs)
+        """Initialize with sha256."""
+        kwargs.update({'shalen': 256})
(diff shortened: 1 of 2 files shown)


autospec/pkg_integrity.py
```diff
@@ -498,8 +499,11 @@ def verify(self):
         package_info = info.get('info', None)
         if package_info is not None:
             EMAIL = package_info.get('author_email', '')
-        self.md5_digest = release_info.get('md5_digest', '')
-        return self.verify_md5()
+        sha256 = ''
+        digests = release_info.get('digests', '')
+        if digests:
+            sha256 = digests.get('sha256', '')
```
bryteise, line 505: Opportunity for `if := digests`

autospec/pkg_integrity.py
```diff
@@ -498,8 +499,11 @@ def verify(self):
         package_info = info.get('info', None)
         if package_info is not None:
             EMAIL = package_info.get('author_email', '')
-        self.md5_digest = release_info.get('md5_digest', '')
-        return self.verify_md5()
+        sha256 = ''
```
bryteise, line 502: The verify function docs still say verify with MD5.

autospec/pkg_integrity.py
phmccarty: good idea, will update

autospec/pkg_integrity.py
phmccarty: thanks, I will update the docstring

PR #765, opened 2022-11-02
Be more aggressive on avx512

fenrus75:


autospec/pkg_integrity.py
```diff
@@ -665,7 +665,7 @@ def get_verifier(filename):
 def get_input(message, default):
     """Parse user input."""
     try:
-        import_key = input(message)
+        import_key = 'y'
```
phmccarty, line 668: I don't think we should auto-accept new pubkey imports...

autospec/specfiles.py
```diff
@@ -671,7 +671,7 @@ def write_license_files(self):
                 file2 = self.hashes[lfile]
                 # Use the absolute path to the source license file b/c we don't know for sure where we are
                 lfile = lfile.replace(self.version, "%{version}")
-                self._write_strip("cp " + "%{_builddir}/" + lfile + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + "\n")
+                self._write_strip("cp " + "%{_builddir}/" + lfile + " %{buildroot}/usr/share/package-licenses/" + self.name + "/" + file2 + " || :\n")
```
phmccarty, line 674: What is the reason for ignoring copy errors?

fenrus75: there's a few packages where the license file is a dead symlink which has no other way for us to work around

phmccarty: We appear to skip the license file for that case... (see https://github.com/clearlinux/autospec/pull/565) I just tested running autospec on a tarball with a broken license symlink, and can confirm that the exception is handled as expected.

Do you remember which packages are affected? I can test those...

PR #800, opened 2023-09-27
Update flag management

bryteise:
Not quite the complete overhaul that is needed (specfile flag generation really could use some unittests) but this change with the macros change in clr-rpm-config makes the flags more consistent.

This is a functional change so expect flags to be different after this change but to my testing there weren't any packages broken.

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -51,8 +51,6 @@ class Specfile(object):
         self.default_desc = ""
         self.locales = []
         self.build_dirs = dict()  # Build directories, indexed by source URL
-        self.need_avx2_flags = False
-        self.need_avx512_flags = False
         self.tests_config = ""
         self.excludes = []
         self.file_maps = {}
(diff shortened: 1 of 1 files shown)


autospec/specfiles.py
```diff
@@ -1008,11 +1005,11 @@ def write_configure_pattern(self):
             self._write_strip("unset PKG_CONFIG_PATH")
             self._write_strip("pushd ../buildavx2/" + self.config.subdir)
             self.write_build_prepend()
-            self._write_strip("export CFLAGS=\"$CFLAGS -m64 -march=x86-64-v3 -Wl,-z,x86-64-v3\"")
-            self._write_strip("export CXXFLAGS=\"$CXXFLAGS -m64 -march=x86-64-v3 -Wl,-z,x86-64-v3\"")
-            self._write_strip("export FFLAGS=\"$FFLAGS -m64 -march=x86-64-v3 -Wl,-z,x86-64-v3\"")
-            self._write_strip("export FCFLAGS=\"$FCFLAGS -m64 -march=x86-64-v3\"")
-            self._write_strip("export LDFLAGS=\"$LDFLAGS -m64 -march=x86-64-v3\"")
+            self._write_strip('CFLAGS="$CLEAR_INTERMEDIATE_CFLAGS -m64 -march=x86-64-v3 -Wl,-z,x86-64-v3 "')
```
fenrus75, line 1008: doesn't this need an export ?

bryteise: Not from my testing looking at the V3 vs 32bit files that I was poking for this change.

PR #801, opened 2023-09-29
Switch to tomllib

bryteise:


bryteise:
Ignoring the lint as it is being replaced in a future clear version.

PR #831, opened 2024-04-11
Add gnupg as a buildreq if we'll need it during build

bwarden:


bwarden:
There may be a _better_ place to do this, but for now it works.

fenrus75 (approved):

Issue #837, opened 2024-06-26
cargo_vendor doesnt account for sources not from crates.io


K1ngfish3r:
Trying to vendor [zed](https://github.com/zed-industries/zed), I came across github sources being vendored as part of the command. When running `cargo vendor > .cargo/config.toml`, it looks something like this
```
[source.crates-io]
replace-with = "vendored-sources"

[source."git+https://github.com/KillTheMule/nvim-rs?branch=master"]
git = "https://github.com/KillTheMule/nvim-rs"
branch = "master"
replace-with = "vendored-sources"

[12 sources]...

[source."git+https://github.com/zed-industries/lsp-types?rev=72357d6f6d212bdffba3b5ef4b31d8ca856058e7"]
git = "https://github.com/zed-industries/lsp-types"
rev = "72357d6f6d212bdffba3b5ef4b31d8ca856058e7"
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
```

Whilst the vendor tarball is created with those deps, Autospec currently doesn't echo in the github sources to account for this, and the builder tries to access the internet.

K1ngfish3r:
I tried v16 and I'm unable to see anything being appended to .cargo/config,toml. I believe this behaviour will also affect other rust packages

K1ngfish3r:
can confirm working with v16 after sync to 42020

PR #839, opened 2024-06-29
Use vendors.txt for appending sources for cargo_vendor builds

K1ngfish3r:
For use with https://github.com/clearlinux/common/pull/20
Closes https://github.com/clearlinux/autospec/issues/837


bryteise:
I wish more rust packages would automate providing their vendor as part of a full source archive release and then we don't need autospec to do anything with vendoring...

Given the world doesn't work that way, using the output of cargo vendor is certainly the best way forward.

K1ngfish3r:
lgtm

PR #842, opened 2024-07-08
Add cargo vendor drop in style file

bryteise:
Add support for cargo_vendors which should be pretty much the output of a 'cargo vendor' run. This will be appended to any existing .cargo/config.toml file in the sources. It is intended to mostly come from the vendor.py in common but useful for other builds as well which don't need to enable cargo_vendor in options.conf.

### DIFF
--- a/autospec/config.py
+++ b/autospec/config.py
@@ -68,6 +68,7 @@ class Config(object):
 
     def __init__(self, download_path):
         """Initialize Default configuration settings."""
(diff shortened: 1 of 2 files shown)


bryteise:
A slight reworking of #839 to avoid adding the vendors.txt file to the other repo while also enabling drop in support for packages that don't use cargo_vendor in options.conf.

bryteise:
Also should fix #837

PR #843, opened 2024-07-22
Fatal log update

bryteise:


autospec/build.py
```diff
@@ -153,17 +154,21 @@ def parse_buildroot_log(self, filename, returncode):
...
             loglines = rootlog.readlines()
         missing_pat = re.compile(r"^.*No matching package to install: '(.*)'$")
         for line in loglines:
             match = missing_pat.match(line)
             if match is not None:
-                util.print_fatal("Cannot resolve dependency name: {}".format(match.group(1)))
-                is_clean = False
-        return is_clean
+                if fatal_buf:
+                    fatal_buf = f"{fatal_buf}\nCannot resolve dependency name: {match.group(1)}"
```
bwarden, line 166: What about making fatal_buf a list instead so we can just append to it cleanly and join with LFs when we print?

fenrus75 (approved):

autospec/build.py
bryteise: Oh nice, thanks!

fenrus75 (approved):

PR #844, opened 2024-07-24
Enable working with Jinja2 spec templates

bryteise:
Allow autospec to handle building packages that use a Jinja2 formatted *.spec.template file. Currently only package_name, package_version, package_release and package_url fields are supported but others can be added as needed (patches and archives are good next steps).

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -26,6 +26,8 @@ import types
 from collections import OrderedDict
 
 import git
+from jinja2 import Environment
+from jinja2.loaders import DictLoader
 from util import _file_write, open_auto
 
 AVX2_CFLAGS = "-march=x86-64-v3"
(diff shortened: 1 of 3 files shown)


autospec/specfiles.py
```diff
@@ -73,12 +75,40 @@ def __init__(self, url, version, name, release, config, requirements, content):
 
     def write_spec(self):
         """Write spec file."""
-        self.specfile = open_auto("{}/{}.spec".format(self.config.download_path, self.name), "w")
+        spec_path = f"{os.path.join(self.config.download_path, self.name)}.spec"
+        self.specfile = open_auto(spec_path, "w")
         self.specfile.write_strip = types.MethodType(_file_write, self.specfile)
 
+        # last change to sanitize url for template and build types
```
bwarden, line 82:
```suggestion
        # last chance to sanitize url for template and build types
```

autospec/specfiles.py
```diff
@@ -73,12 +75,40 @@ def __init__(self, url, version, name, release, config, requirements, content):
...
+        if self.config.urlban:
```
bwarden, line 83: Why does urlban become optional?

autospec/specfiles.py
```diff
@@ -73,12 +75,40 @@ def __init__(self, url, version, name, release, config, requirements, content):
...
+            clean_url = re.sub(self.config.urlban, "localhost", self.url)
+            # Duplicate prefixes entry before we change the url
+            self.content.prefixes[clean_url] = self.content.prefixes.get(self.url)
+            self.url = clean_url
+
+        template_path = f"{spec_path}.template"
+
+        if os.path.isfile(template_path):
+            # make templates have a template build pattern
+            self.config.default_pattern = "template"
```
bwarden, line 93: Will there be a `write_template_pattern()` to implement this?

autospec/specfiles.py
```diff
@@ -109,6 +139,7 @@ def write_spec(self):
         self.write_lang_files()
 
         self.specfile.close()
+        return "generate"
```
bwarden, line 142: Does "generate" mean continue with today's behavior of autospec -- generating the specfile from scratch?

autospec/specfiles.py
bryteise: I didn't think this changed anything logically, it just moved it from the nvr writing function. The reasoning for this to be optional is because the config for urlban isn't always existing (external builders probably don't use that config).

autospec/specfiles.py
bryteise: Right now the second if isfile block is effectively that at the moment.

The template pattern at this point doesn't share anything outside the header with the normal flow though so I didn't want to make it seem too similar to the other patterns which are more focused on %build and %install sections.

autospec/specfiles.py
bryteise: Yep, open to other words/handling. I'm just treating it as an enum basically.

autospec/specfiles.py
bwarden: OK, I guess we're reading it from common, not from the package options.conf.

Issue #855, opened 2024-10-26
build_pattern autogen missing subdir routine


K1ngfish3r:
it missing

PR #856, opened 2024-10-26
add subdir for autogen

K1ngfish3r:
fixes https://github.com/clearlinux/autospec/issues/855

### DIFF
--- a/autospec/specfiles.py
+++ b/autospec/specfiles.py
@@ -25,9 +25,10 @@ import time
 import types
 from collections import OrderedDict
 
-import git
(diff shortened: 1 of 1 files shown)


bryteise:
flake didn't like you moving that git import

K1ngfish3r:
buh why? I used `isort`

bryteise:
Some difference in setting perhaps? isort on clear linux gives me the same output as the current import layout.

K1ngfish3r:
fresh install btw
image

PR #860, opened 2024-11-12
Add handling for emacs jit files

bryteise:
emacs jit files are dynamically generated names that are under a path like /usr/libexec/emacs/$version/$arch/emacs-$hash.pdmp. Collect these (though there is only 1 per build at this time) files under a single glob to avoid needing to account for the filename changing between builds.


bwarden:
+1

bwarden (approved):

bryteise:
Eh actually this is better with an emacs patch it seems.

PR #871, opened 2025-04-02
Catch and ignore gpg errors on old-format ASCII messages

bwarden:
Apparently GPG changed the preferred format for ASCII messages such as ascii-armored signatures. When parsing the older format, gpg returns 2 and emits the following message to STDERR:

gpg: no valid OpenPGP data found.

BUT it may still process the data appropriately.

We'll look for this case specifically, and continue processing as if all is well -- if the signature file is actually bad, the whole verification process should fail in different ways.


autospec/pkg_integrity.py
```diff
@@ -113,7 +113,9 @@ def verify(self, _, tarfile, signature):
             # sig file may be ascii-armored, so dearmor it first...
             args = self.args + ['--dearmor', '--output', '-', signature]
             output, err, code = self.exec_cmd(args)
-            if code != 0:
+            if code == 2 and err.decode('utf-8').startswith('gpg: no valid OpenPGP data found.'):
+                util.print_warning("Ignoring: {}".format(err.decode('utf-8')))
+            elif code != 0:
```
bryteise, line 118: Rather than decoding the err twice, mind just doing a `errmsg = err.decode('utf-8')` after exec_cmd (my own minor annoyance, feel free to ignore).

autospec/pkg_integrity.py
```diff
@@ -477,7 +479,8 @@ def verify(self, recursion=False):
         if sign_isvalid(self.package_sign_path) is False:
             self.print_result(False, err_msg='{} is not a GPG signature'.format(self.package_sign_path))
             try:
-                os.unlink(self.package_sign_path)
+                #os.unlink(self.package_sign_path)
```
bryteise: Any reason to keep the try and code commented out? Might as well just print and return.

bwarden: That was for debugging so it would stop pulling the rug out from under me :) Restored the unlink so it will still happen when the signature really fails.

autospec/pkg_integrity.py
```diff
@@ -625,9 +628,12 @@ def parse_gpg_packets(filename, verbose=True):
...
     args = ["gpg", "--list-packets", filename]
     try:
-        out, err = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE).communicate()
-        if err.decode('utf-8') != '' and verbose is True:
-            print(err.decode('utf-8'))
+        process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, err = process.communicate()
+        if process.returncode == 2 and err.decode('utf-8').startswith('gpg: no valid OpenPGP data found.'):
+            util.print_warning("Ignoring: {}".format(err.decode('utf-8')))
+        elif err.decode('utf-8') != '' and verbose is True:
```
bryteise, line 634: Similar comment around using an errmsg variable as above.

autospec/pkg_integrity.py
bwarden: Yeah, worth cleaning up.

autospec/pkg_integrity.py
bwarden: will fix here too

bwarden:
Underlying issue is a bug in gnupg. Fixing that via backport instead makes this error case go away.

